PR #100, opened 2021-03-14
BUGFIX: Pass shadow-specific per-view descriptor set.

DavidVonDerau:
Commit 5637071b4ecb10f19b9b6e42def1d3071ef160cb wasn't passing the per-view descriptor sets to the shadow pass and no shadows were drawn as a result.

It didn't make sense to share the same per-view uniform as the opaque pass because the shadow pass doesn't need any of the lighting data, so this commit defines new shadow uniforms in `mesh_shadow_map.glsl`. As there's no usage of most of the definitions in `mesh.glsl`, I removed the import of that entirely. This greatly reduced the size of the generated code for `mesh_shadow_map.vert`. In the demo's `Mesh` feature, `per_view_descriptor_set` is no longer an `Option` because it's required for both the opaque & the shadow draw calls. The shadow map view data is collected in `shadow_per_view_descriptor_set_layouts`, populated by the `"create per view dsecriptor sets"` scope, and ultimately assigned to each node at the `add_render_node` call-site. The `is_shadow_pass` argument was removed from `add_render_node` because it wasn't used anymore.


demo/shaders/glsl/mesh_shadow_map.glsl
```diff
@@ -0,0 +1,40 @@
...
+//         mag_filter: Linear,
+//         min_filter: Linear,
+//         mip_map_mode: Linear,
+//         address_mode_u: Repeat,
+//         address_mode_v: Repeat,
+//         address_mode_w: Repeat,
+//         max_anisotropy: 16.0,
+//     )
+// ])]
+layout (set = 0, binding = 1) uniform sampler smp;
```
DavidVonDerau, line 19: It's not clear to me if this needs to be defined for this shadow. We only bind a depth attachment, right?

demo/shaders/glsl/mesh_shadow_map.glsl
```diff
@@ -0,0 +1,40 @@
...
+//         mip_map_mode: Linear,
+//         address_mode_u: ClampToEdge,
+//         address_mode_v: ClampToEdge,
+//         address_mode_w: ClampToEdge,
+//         anisotropy_enable: true,
+//         max_anisotropy: 16.0,
+//         compare_op: Greater,
+//     )
+// ])]
+layout (set = 0, binding = 2) uniform sampler smp_depth;
```
DavidVonDerau, line 34: I wondered if I should pull out these sampler definitions into a common file other than `mesh.glsl` so they could be imported into both `mesh.frag` and `mesh_shadow_map.vert`, but I wasn't sure if any code depended on these samplers being the same between both shaders. It seemed like the answer would be no.

demo/shaders/glsl/mesh_shadow_map.glsl
aclysma: Yeah, I don't think we would need that now

demo/shaders/glsl/mesh_shadow_map.glsl
aclysma: Yeah, I don't think we need any samplers when rendering shadow maps

demo/shaders/glsl/mesh_shadow_map.glsl
DavidVonDerau: Removed.

demo/shaders/glsl/mesh_shadow_map.glsl
DavidVonDerau: Removed.

PR #102, opened 2021-03-15
Add depth prepass

DavidVonDerau:
This change removes the shadow pass from the `mesh.material` and instead defines a new `depth.material`. The `MeshAssetPart` created by `gltf/assets.rs` now have boolean values for `casts_shadows` or `receives_shadows` instead of the optional shadow material -- this assumes that all generating shadows for a given mesh part can use the new predefined material. The new `depth.material` is passed through the `mesh/extract` job into `mesh/prepare` where it is used for both shadows and the depth pre-pass.


demo/src/assets/gltf/assets.rs
```diff
@@ -138,8 +138,8 @@ pub struct MeshAssetData {
 pub struct MeshAssetPart {
     pub opaque_pass: MaterialPass,
     pub opaque_material_descriptor_set: DescriptorSetArc,
-    // These are optional because we might want to disable casting shadows
-    pub shadow_map_pass: Option<MaterialPass>,
+    pub casts_shadows: bool,
+    pub receives_shadows: bool,
```
DavidVonDerau, line 142: These `casts_shadows` or `receives_shadows` can probably just be removed if the plan is to move them to some higher level `RenderProxy` concept.

rafx-assets/src/assets/asset_manager.rs
```diff
@@ -279,6 +279,25 @@ impl AssetManager {
             .map(|x| x.material_pass_resource.clone())
     }
 
+    pub fn get_material_pass_by_name(
```
aclysma, line 282: I've been thinking about having fewer functions (i.e. eventually removing accessors like get_material_pass_by_index) on AssetManager. Maybe functions like this could be on the MaterialAsset and we just have a single function returning the &MaterialAsset. So someone could do `asset_manager.committed_asset(material_handle)?.get_material_pass_by_phase(...)`

demo/src/assets/gltf/assets.rs
aclysma: Do you mean as fields in `MeshRenderNode` or something else?

DavidVonDerau: Yeah, something populated by the extract job. I can see a reason why you'd import a model into the world and then, on a per "entity" basis, decide whether or not it should cast shadows with some component. During the extract, at the same time we pull out the model info, we'd also check that component to see if we should cast or receive shadows.

demo/src/features/mesh/extract.rs
```diff
@@ -122,7 +123,20 @@ impl ExtractJob for MeshExtractJob {
...
+        let static_resources = extract_context
+            .render_resources
+            .fetch::<MeshStaticResources>();
+
+        let asset_manager = extract_context
+            .render_resources
+            .fetch::<AssetManagerRenderResource>();
+
+        let depth_material = asset_manager
+            .get_material_pass_by_name(&static_resources.depth_material, "Depth")
```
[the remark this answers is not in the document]
DavidVonDerau: I was torn on that. I felt that the index lookup might be more brittle (e.g. if we add a multi-view depth pass for point lights) but obviously right now it doesn't matter since there's only the one pass.

rafx-assets/src/assets/asset_manager.rs
DavidVonDerau: That's fine with me. I'll move them.

demo/assets/materials/depth.material
```diff
@@ -0,0 +1,19 @@
...
+        (
+            name: Some("Depth"),
+            fixed_function_state: (
+                depth_testing: EnabledReverseZ,
+                cull_mode: Some(None),
+            ),
+            shaders: [
+                (
+                    stage: Vertex,
+                    shader_module: "../shaders/mesh_shadow_map.vert.cookedshaderpackage",
```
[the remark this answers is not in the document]
DavidVonDerau: Yep. Will do.

demo/src/features/mesh/extract.rs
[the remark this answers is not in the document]
aclysma: We could consider a function like `get_only_material_pass` (or some other better name!) that asserts there is exactly 1 material pass in the material and returns it. That would make intent clear and avoid silent failure of using the wrong material pass.

demo/src/assets/gltf/assets.rs
aclysma: Yeah that sounds good!

demo/src/features/mesh/extract.rs
```diff
@@ -122,7 +123,21 @@ impl ExtractJob for MeshExtractJob {
             })
             .collect();
 
+        let static_resources = extract_context
+            .render_resources
+            .fetch::<MeshStaticResources>();
+
+        let asset_manager = extract_context
```
aclysma, line 130: We fetched this previously so this line can probably be removed

demo/src/features/imgui/extract.rs
```diff
@@ -59,7 +59,8 @@ impl ExtractJob for ImGuiExtractJobImpl {
             .fetch::<ImguiStaticResources>()
             .imgui_material;
         let imgui_material_pass = asset_manager
```
aclysma, line 61: Is it possible to do:

```rust
let imgui_material_pass = asset_manager
             .get_material_pass_by_index(imgui_material, 0)
             .committed_asset(imgui_material)
             .unwrap()
             .get_single_material_pass()
             .unwrap();
```

demo/src/features/debug3d/extract.rs
```diff
@@ -38,7 +38,8 @@ impl ExtractJob for Debug3dExtractJob {
             .debug3d_material;
 
         let debug3d_material_pass = asset_manager
```
aclysma, line 40: Can we do

```rust
        let debug3d_material_pass = asset_manager
             .get_material_pass_by_index(&debug3d_material, 0)
             .committed_asset(&debug3d_material)
             .unwrap();
             .get_single_material_pass()
             .unwrap();
```

demo/src/features/text/extract.rs
```diff
@@ -37,7 +37,8 @@ impl ExtractJob for TextExtractJob {
             .fetch::<TextStaticResources>()
             .text_material;
         let text_material_pass = asset_manager
-            .get_material_pass_by_index(&text_material, 0)
+            .committed_asset(&text_material)
+            .map(|x| x.get_single_material_pass().unwrap())
```
aclysma, line 41: same as other comment, might not need the map()

demo/src/render_graph_generator/mod.rs
```diff
@@ -125,19 +125,23 @@ impl RenderGraphGenerator for DemoRenderGraphGenerator {
             compute_test::compute_test_pass(&mut graph_context, &compute_test_pipeline);
 
         let bloom_extract_material_pass = asset_manager
-            .get_material_pass_by_index(&static_resources.bloom_extract_material, 0)
+            .committed_asset(&static_resources.bloom_extract_material)
+            .map(|x| x.get_single_material_pass().unwrap())
```
aclysma, line 131: these as well

rafx-assets/src/assets/graphics_pipeline/assets.rs
```diff
@@ -349,6 +350,51 @@ impl MaterialAsset {
     ) -> Option<usize> {
         self.inner.pass_phase_to_index.get(&index).copied()
     }
+
+    pub fn get_single_material_pass(&self) -> RafxResult<ResourceArc<MaterialPassResource>> {
+        if self.inner.passes.len() == 1 {
+            Ok(self
+                .inner
+                .passes
```
aclysma, line 370: since we checked length, we could do inner.passes[0].material_pass_resource

rafx-assets/src/assets/graphics_pipeline/assets.rs
```diff
@@ -349,6 +350,51 @@ impl MaterialAsset {
...
+                .get(0)
+                .unwrap()
+                .material_pass_resource
+                .clone())
+        } else {
+            Err(RafxError::StringError(
```
aclysma, line 376: This could probably be Err("Found more than one MaterialPass in MaterialAsset.")?

I'd mention "get_single_material_pass" in the error. Something like "get_single_material_pass was called on a MaterialAsset that does not have exactly 1 MaterialPass"

rafx-assets/src/assets/graphics_pipeline/assets.rs
```diff
@@ -349,6 +350,51 @@ impl MaterialAsset {
...
+    }
+
+    pub fn get_material_pass_by_index(
+        &self,
+        index: usize,
+    ) -> Option<ResourceArc<MaterialPassResource>> {
+        self.inner
+            .passes
+            .get(index)
+            .map(|x| x.material_pass_resource.clone())
```
aclysma, line 389: Could probably do .get(index)?.clone()

aclysma: same for the two similar cases below

DavidVonDerau: This did not work, but I addressed everything else.

aclysma:
Nice, I'll merge it once CI finishes :)

PR #104, opened 2021-03-16
Move render nodes to render resources, fix sprite rendering

aclysma:


demo/shaders/glsl/sprite.glsl
```diff
@@ -9,9 +9,9 @@ layout(set = 0, binding = 0) uniform Args {
 //             mag_filter: Linear,
 //             min_filter: Linear,
 //             mip_map_mode: Linear,
-//             address_mode_u: Repeat,
-//             address_mode_v: Repeat,
-//             address_mode_w: Repeat,
+//             address_mode_u: Mirror,
```
aclysma, line 12: Repeat was causing the color of the edge of one side of the texture to show up on the other side

demo/src/components.rs
```diff
@@ -8,8 +8,8 @@ use rafx::visibility::DynamicAabbVisibilityNodeHandle;
 
 #[derive(Clone)]
 pub struct MeshComponent {
-    pub render_node: MeshRenderNodeHandle,
-    pub visibility_node: DynamicAabbVisibilityNodeHandle,
+    pub render_node: Option<MeshRenderNodeHandle>,
```
aclysma, line 11: these are now populated during the extract phase. this does break the general convention that extract should be read-only over all data so that we can do it in parallel across all data. not sure what to do about it

demo/src/features/mesh/plugin.rs
```diff
@@ -32,16 +32,21 @@ impl RendererPlugin for MeshRendererPlugin {
         _upload: &mut RafxTransferUpload,
     ) -> RafxResult<()> {
         render_resources.insert(ShadowMapResource::default());
+        render_resources.insert(MeshRenderNodeSet::default());
+
+        render_resources.ensure_exists_or_add_default::<StaticVisibilityNodeSet>();
```
aclysma, line 37: the ensure_exists_or_add_default call is so that if two plugins require the same thing to be created, they can both call this and not stomp on each other

demo/src/features/sprite/prepare.rs
```diff
@@ -120,13 +120,13 @@ impl PrepareJob for SpritePrepareJob {
...
         let half_height = 400.0 / aspect_ratio;
-        let view_proj = glam::Mat4::orthographic_rh_gl(
+        let view_proj = glam::Mat4::orthographic_rh(
             -half_width,
             half_width,
             -half_height,
             half_height,
-            -100.0,
-            100.0,
+            1000.0,
```
aclysma, line 128: we're using reverse z, so flip the sign here. I also think for sprite rendering, depth: [-1000,1000] is plenty of precision and should be enough for most people

rafx-base/src/resource_map.rs
```diff
@@ -107,6 +107,15 @@ impl ResourceMap {
         }
     }
 
+    pub fn ensure_exists_or_add_default<R>(&mut self)
```
DavidVonDerau, line 110: I would rename this to `try_insert` with signature 
```rust
pub fn try_insert<R>(
        &mut self,
        r: R,
    ) -> bool
```

Then it would be called like

```rust
        render_resources.insert(MeshRenderNodeSet::default());

        render_resources.try_insert(StaticVisibilityNodeSet::default());
```

The difference being that `insert` would explode on overwrite (as implied by the `TODO` in `insert_by_id`) whereas `try_insert` would only insert the value if it wasn't already present. I think that's a little bit cleaner because it keeps the syntax the same between both calls -- the only difference is the function name.

aclysma: I think for this particular use case (multiple plugins wanting to set the value to something) we should only allow creating the default object. This way the order that plugins are created doesn't affect whether we insert what the mesh plugin's desired state or the sprite plugin's desired state. Maybe in some circumstance we would want that, in which case the plugin should probably insert and verify nothing else was inserted first. Also it's possible an end-user might pre-populate render resources with an overridden instance of type T as a way of configuring something. (Bevy uses this for configuring the task system for example.)

I like the try_insert(), so maybe we could to try_insert_default(), where it inserts T::default() rather than taking a T. (well, R I guess :D)

DavidVonDerau (approved):
Looks fine, other than the one nit about maybe changing `ensure_exists_or_add_default` to `try_insert` to make it closer in usage.

rafx-base/src/resource_map.rs
DavidVonDerau: `try_insert_default<T>` sounds good.

demo/src/components.rs
aclysma: We discussed this on discord and considered a few options:
1) Rely on jobs that write to the same components do not run at the same time (this is checked at runtime)
2) Use interior mutability. Probably not better than option 1 because it just silently hides the write and could result in inconsistent logical behavior depending on order of operations. But it's an option.
3) Defer updating these fields until later. This add some complexity and potential for error (i.e. creating a render node and setting the ID on the component is no longer an atomic operation.)

For now the current implementation is most like option 1. We'll go with it for now and maybe a better solution will present itself later.

PR #109, opened 2021-03-21
Remove add_render_node_reservations

aclysma:
This removes a callback on plugins that will probably seem strange to end-users. The performance benefit of pre-sizing the vector is not much (0.01ms if that) and I think it would be more accurately implemented anyways by looking at how many nodes were used in the previous frame + some extra %.


DavidVonDerau (approved):
The CI job failed but the changes look good if you can resolve that.

PR #112, opened 2021-03-28
Check out as LF on windows

kabergstrom:
Fixes false changes with shader-processor generated code since it always writes LF while git checks out the generated files as CRLF.

based off #111


aclysma:
Will try to fix the shader processor to output CRLF on windows since changing .gitattributes would not solve this for people working in their own repos.

PR #122, opened 2021-04-09
Feature clean-up

DavidVonDerau:
This PR attempts to establish a new consistent format for defining features.

- Introduce new `use` preludes for quickly defining the `use` statements required by the extract, prepare, write, and renderer traits.
- The `declare_render_feature` macro now defines `render_feature_index()` and `render_feature_debug_name()` functions so that the feature code doesn't keep repeating the name of the feature type.
- The `declare_render_feature` macro now defines &str names to use for the profiling::scope! macro.
- All extract / prepare / write jobs are assumed to follow the `ExtractJob`, `PrepareJob`, `WriteJob` naming scheme.
- Jobs are defined in a `jobs` directory.

In general, data definitions are scoped as narrowly as possible.

- Prepared data is defined inside of the `write` file. The reason for defining the data in `write` instead of `prepare` is because ultimately the `writer` is the _consumer_ of that data -- the prepared data is more or less an interface that the `writer` is asking the `prepare` job to produce. 
- Vertex layouts are defined inside of the `write` file.
- Extracted data is defined inside of the `prepare` file. The same rationale for prepared data applies here -- the `prepare` job is defining an interface (the extracted data) that the `extract job` must meet.
- Jobs are always constructed by `new` and never by `{ ...fields... }` initialization.
- Pass-through structs are avoided. If a job would create a struct just for that to become a member on the next job, the struct is gone and the members have been moved onto the job impl.
- Jobs can construct the writer earlier in the `prepare` function so that they can use functions on the `writer`.

Furthermore
- If a feature exposes structs for use outside of the jobs, those should go into the top-level folder and be re-exported inside of the `mod`.
- If a feature needs to define common functionality for itself that doesn't fit into any of the existing jobs, put that into an `internal` folder and don't export it outside of the feature.

Finally
- Views can now filter out features using a `RenderFeatureMask` in the same way as they can filter on phases with the `RenderPhaseMask`. When building the frame packet, we'll filter out visible nodes that don't match the view's feature mask. The `sprite_scene` shows an example of disabling the skybox by not registering the main view with that feature.
- While on the shadow scene, it's possible to show or hide the debug feature, the text feature, the skybox feature, or the shadows.

image


demo/src/features/debug3d/extract.rs
```diff
@@ -1,28 +1,25 @@
...
+rafx::declare_render_feature_extract_job!();
+
+use super::{DebugDraw3DResource, StaticResources};
 use rafx::assets::AssetManagerRenderResource;
-use rafx::nodes::{
-    ExtractJob, FramePacket, PrepareJob, RenderFeature, RenderFeatureIndex,
-    RenderJobExtractContext, RenderView,
-};
 
-pub struct Debug3dExtractJob {}
```
aclysma, line 10: I think we can remove the "Impl" suffix. It's left over from when there was another layer to the jobs that had the destiny-style callbacks.

I think we should keep the prefix (Debug3d in this case) so that we don't have N structs with the same name.

demo/src/features/debug3d/extract.rs
```diff
@@ -34,7 +31,7 @@ impl ExtractJob for Debug3dExtractJob {
 
         let debug3d_material = &extract_context
             .render_resources
-            .fetch::<Debug3DStaticResources>()
+            .fetch::<StaticResources>()
```
aclysma, line 34: Would keep the Debug3D prefix

demo/src/features/debug3d/extract.rs
```diff
@@ -1,28 +1,25 @@
-use crate::features::debug3d::plugin::Debug3DStaticResources;
-use crate::features::debug3d::prepare::Debug3dPrepareJobImpl;
-use crate::features::debug3d::{Debug3dRenderFeature, DebugDraw3DResource, ExtractedDebug3dData};
+rafx::declare_render_feature_extract_job!();
```
aclysma, line 1: I don't want to use macros to add use statements. An alternative would be to do something like:

```
pub mod job_prelude {
    pub use crate::features::etc;
}
```

and then use crate::job_prelude::*;

demo/src/features/debug3d/extract.rs
```diff
@@ -1,28 +1,25 @@
...
 
-impl ExtractJob for Debug3dExtractJob {
+impl ExtractJob for ExtractJobImpl {
     fn extract(
         self: Box<Self>,
         extract_context: &RenderJobExtractContext,
         _frame_packet: &FramePacket,
         _views: &[RenderView],
     ) -> Box<dyn PrepareJob> {
-        profiling::scope!("Debug3d Extract");
```
aclysma, line 25: I kind of prefer to just use the string directly. Every time we add an indirection to something, it means someone has to go look somewhere else in the code to figure out what's going on, and the value indirecting here brings is minimal.

demo/src/features/debug3d/mod.rs
```diff
@@ -1,53 +1,71 @@
...
+
 mod extract;
-mod plugin;
 mod prepare;
 mod write;
-pub use plugin::Debug3DRendererPlugin;
 
-pub use debug3d_resource::*;
-use rafx::api::RafxPrimitiveTopology;
+mod public;
```
aclysma, line 10: if something is public I would just leave it in the root of the feature's module.

demo/src/features/debug3d/mod.rs
```diff
@@ -1,53 +1,71 @@
...
-pub type Debug3dUniformBufferObject = shaders::debug_vert::PerFrameUboUniform;
+pub use public::debug3d_resource::DebugDraw3DResource;
 
-/// Vertex format for vertices sent to the GPU
-#[derive(Clone, Debug, Copy, Default)]
-#[repr(C)]
-pub struct Debug3dVertex {
-    pub pos: [f32; 3],
-    pub color: [f32; 4],
+struct StaticResources {
```
aclysma, line 17: Would call it Debug3dStaticResources so we don't have N StaticResources structs

demo/src/features/debug3d/mod.rs
```diff
@@ -1,53 +1,71 @@
...
+    pub debug3d_material: Handle<MaterialAsset>,
 }
 
-lazy_static::lazy_static! {
-    pub static ref DEBUG_VERTEX_LAYOUT : VertexDataSetLayout = {
-        use rafx::api::RafxFormat;
+pub struct RendererPluginImpl;
+
+impl RendererPlugin for RendererPluginImpl {
```
aclysma, line 23: Debug3dRendererPlugin

demo/src/features/debug3d/mod.rs
```diff
@@ -1,53 +1,71 @@
...
+    fn configure_render_registry(
+        &self,
```
aclysma, line 25: I preferred the plugin in a separate file

demo/src/features/debug3d/prepare.rs
```diff
@@ -1,51 +1,57 @@
...
 use rafx::framework::{MaterialPassResource, ResourceArc};
-use rafx::nodes::{
-    FeatureCommandWriter, FeatureSubmitNodes, FramePacket, PrepareJob, RenderFeature,
-    RenderFeatureIndex, RenderJobPrepareContext, RenderView, ViewSubmitNodes,
-};
 
-pub struct Debug3dPrepareJobImpl {
+pub type Debug3dUniformBufferObject = shaders::debug_vert::PerFrameUboUniform;
+
+pub struct PrepareJobImpl {
```
aclysma, line 12: If we rename, would rename to Debug3dPrepareJob

demo/src/features/debug3d/write.rs
```diff
@@ -1,25 +1,101 @@
...
+rafx::declare_render_feature_write_job!();
+
+use rafx::api::RafxPrimitiveTopology;
+use rafx::api::RafxVertexBufferBinding;
+use rafx::framework::{VertexDataLayout, VertexDataSetLayout};
+
+/// Vertex format for vertices sent to the GPU
+#[derive(Clone, Debug, Copy, Default)]
+#[repr(C)]
+pub struct Debug3DVertex {
```
aclysma, line 10: Since this is being used from the prepare job and the write job, I would move it to the base mod.rs - it's doesn't clearly "belong" to either one exclusively

demo/src/features/debug3d/write.rs
```diff
@@ -1,25 +1,101 @@
...
+}
+
+pub struct FeatureCommandWriterImpl {
+    vertex_buffer: Option<ResourceArc<BufferResource>>,
+    draw_calls: Vec<Debug3DDrawCall>,
+    debug3d_material_pass: ResourceArc<MaterialPassResource>,
+    per_view_descriptor_sets: Vec<Option<DescriptorSetArc>>,
 }
 
-impl FeatureCommandWriter for Debug3dCommandWriter {
```
aclysma, line 16: Maybe Debug3dWriteJob would be more consistent with the extract and prepare job naming. (In fact you named one of the macros "declare_render_feature_write_job!()"

demo/src/features/debug3d/write.rs
```diff
@@ -66,14 +144,15 @@ impl FeatureCommandWriter for Debug3dCommandWriter {
                 command_buffer.cmd_draw(draw_call.count as u32, draw_call.first_element as u32)?;
             }
         }
+
         Ok(())
     }
 
     fn feature_debug_name(&self) -> &'static str {
-        Debug3dRenderFeature::feature_debug_name()
+        render_feature_debug_name()
```
aclysma, line 152: As a matter of style I would write `super::render_feature_debug_name()` (or `CrateName::function()` if it's wasn't in super) to make it clear to a reader where the function is from.

demo/src/features/skybox/extract.rs
```diff
@@ -0,0 +1,51 @@
+rafx::declare_render_feature_extract_job!();
+
+use super::StaticResources;
```
aclysma, line 3: Drawing the skybox as a feature made several dozen lines into several hundred lines. I have a hard time agreeing with a change that does the same thing but is ~10x the amount of code. I wouldn't do it for example/teaching purposes either because I would not recommend someone to do it this way.

We could add a new resource that the old code fetches to drive whether the skybox is enabled and what material to draw. This would avoid anything skybox-specific being part of the interface for the graph building. In general this is how I would pass arbitrary configuration from user's high-level game logic into the render graph. (We can also use this method for configuring cameras/viewports.)

aclysma: BTW the "plugin" system is very 1.0, minimum viable product. The function calls that it has are simply what was needed to decouple the Renderer from the demo code and push it up into a crate. There may be better ways to provide the same hooks. (like adding resources that contain data.)

demo/src/init.rs
```diff
@@ -53,15 +52,15 @@ pub fn rendering_init(
...
-    resources.insert(MeshRenderNodeSet::default());
-    resources.insert(TileLayerRenderNodeSet::default());
     resources.insert(StaticVisibilityNodeSet::default());
     resources.insert(DynamicVisibilityNodeSet::default());
-    resources.insert(DebugDraw3DResource::new());
-    resources.insert(TextResource::new());
     resources.insert(ViewportsResource::default());
-    resources.insert(TileLayerResource::default());
+
+    mesh::legion_init(resources);
```
aclysma, line 59: I wonder if it would be possible for a user to provide a closure that takes a resource and inserts it into the resources. (We can investigate that some other time, this change is already huge.)

rafx-framework/src/nodes/macro_render_feature.rs
```diff
@@ -31,5 +44,75 @@ macro_rules! declare_render_feature {
...
+        }
+
+        fn render_feature_debug_name() -> &'static str {
+            $struct_name::feature_debug_name()
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! declare_render_feature_mod {
```
aclysma, line 59: I would remove all the macros that do nothing but add use statements. (I actually think it's bad practice for a macro to "leak" use statements). Left an alternative in another comment.

aclysma: Also in general, we could consider having a prelude for the different crates or add pub use in key places so people can do use rafx::assets::prelude::* and get sensible imports.

aclysma (commented):
First pass review. I think we should iterate on changing just one feature for now until we agree on a pattern for the rest to follow.

demo/src/features/debug3d/write.rs
DavidVonDerau: I put this into the `write.rs` file because it's part of the the writer's interface. The reason why the `prepare` job uses it is because the `prepare` job is fulfilling that interface.

image

demo/src/features/debug3d/extract.rs
DavidVonDerau: The benefit of the indirection is that this is just boilerplate that needs to be changed every time you work on a feature. It's easy to forget it, or to not keep it up to date if we change the format of it in same way. With this short-hand, it's N less things to need to remember to change when you copy / paste a new feature.

As for needing to look elsewhere in the code -- if you have any type of editor or IDE that desugars Rust (e.g. for type inference), this shouldn't be hidden at all.

image

demo/src/features/debug3d/prepare.rs
DavidVonDerau: Resolving to consolidate discussion of job names / prefixes into https://github.com/aclysma/rafx/pull/122#discussion_r610704300.

demo/src/features/debug3d/mod.rs
DavidVonDerau: Resolving to consolidate discussion of prefixes into https://github.com/aclysma/rafx/pull/122#discussion_r610704654.

demo/src/features/debug3d/extract.rs
DavidVonDerau: I updated the comment on the `declare_render_feature` macro so that it'll be recognized as an actual doc comment (`///` instead of `//`).

image

demo/src/features/debug3d/mod.rs
DavidVonDerau: I consolidated the plugin / feature definition into the `mod` file because it provides a single location to see the following:

1. The `public` feature declaration itself, e.g. `rafx::declare_render_feature!(ImGuiRenderFeature, DEBUG_3D_FEATURE_INDEX);`
2. Any `public` structs that have been re-exported, e.g. `pub use public::Sdl2ImguiManager;`
3. The static resources the feature expects.
4. The `public` plugin itself, which is really the _main_ interface, and defines the `init` & `destroy` code. (possibly referencing the internal or public structs from 2 & shoving them into `Resources`), including the `initialization` of the static resources from 3.

demo/src/features/skybox/extract.rs
DavidVonDerau: The feature approach has more lines, yes. The actual _work_ (static resources, extract, prepare, draw) is more or less the same, it's just that the trait implementations of the plugin & jobs adds the overhead. However, there are 3 mitigating factors that I think are relevant:

1. Once you've seen one feature's boilerplate, you've seen them all. 

It's not going to be difficult for someone to zero in on the skybox code -- static resources are defined in `mod.rs`, the retrieval from the asset manager is in `extract`, dynamic descriptor sets are in `prepare`, and the submit calls are in `write`. 

The previous code required one to understand that the skybox materials are defined in `DemoStaticResources`,  they're retrieved from the asset manager in `render_graph_generator/mod.rs`, and the call to prepare descriptor sets & submit happens simultaneously in the middle of `opaque_pass.rs` between `OpaqueRenderPhase` and `TransparentRenderPhase` and is delegated to  a file `skybox.rs` in `features` but is not actually a feature.

2. There is a strong benefit in my mind towards reducing the number of different ways to configure a thing.

Consider the skybox. Implemented as a feature, it behaves the same as anything else. You can create a `RenderView` that selectively masks for it -- like a main player view (skybox enabled), a rear-view mirror in a car (skybox enabled), or a minimap (skybox disabled).

If it's not implemented as a feature, but is instead a resource of some type, we then need code to handle different resources per view. So we can put that onto the `RenderView` (e.g. `skybox_enabled`) and the render graph can iterate over all of the views and check if they have the `skybox_enabled` flag set, but at some point it's just _another_ way to describe "this is a thing that may or may not be drawn for a particular view". We'll still end up with dozens of more lines of code, but instead of it following the same pattern as all of the other features commands, it'll be different.

3. I think a large part of your aversion to making the skybox is a feature is that it's "just" a triangle.
 
I can easily imagine a higher fidelity skybox with e.g. multiple layers (to simulate parallax of things like clouds or stars), or an animated texture instead of static, or more complicated shader effects. In my mind, the render graph is specifically for compositing passes & screen-space post-processing effects together in a way that can efficiently manage GPU resources, whereas features are for what is drawn inside of passes or as a per-view post-processing effect. For example, in this slide from the Destiny talk, the render graph feels like the equivalent of Destiny's _"render passes"_ -- shading, tone mapping & resolve, etc. The skybox is a _"render stage"_ -- it's _"execute submission for a specific view for a specific stage of the frame pipeline where we might have content-driven data"_ 

image

demo/src/features/debug3d/extract.rs
DavidVonDerau: I felt that the name prefix was unnecessary because 1.) from a code organization POV, it's already namespaced by the feature folders, and 2.) it's not accessible outside of the feature folder. No external code can try to `use` or `new` up one of these structs.

Connected to the above point, if there's no prefix, there needs to be some type of suffix so it doesn't conflict with the trait name. If we _do_ put the prefix back, then I agree the suffix can be removed.

demo/src/features/debug3d/extract.rs
DavidVonDerau: If `StaticResources` are intended to be public, then I agree and I'll add the prefixes back. 
If they are not intended to be public, this discussion should resolve according to https://github.com/aclysma/rafx/pull/122#discussion_r610704300.

demo/src/features/debug3d/mod.rs
DavidVonDerau: I think the template is easier to understand with the `public` folder because you can re-export that entire directory inside of `mod.rs` with just `pub use public::*`, but you can't do that inside of `mod.rs` without it.

demo/src/features/debug3d/extract.rs
DavidVonDerau: Per discussion offline, will add the prefix & remove the suffix.

demo/src/features/debug3d/extract.rs
DavidVonDerau: Per discussion offline, will add the prefix & remove the suffix.

demo/src/features/debug3d/write.rs
aclysma: We have cases where the Vertex is used by an importer (see `TileLayerVertex` and `MeshVertex`). So I still think that it should be a top-level object in the feature in the mod.rs.

demo/src/features/debug3d/mod.rs
aclysma: When a mod.rs file gets to be 100+ lines, I start looking for ways to pull larger chunks of code out of it. The mesh plugin is already close to 100 lines by itself. In general the plugin seems like a piece that could grow. (pub(super) is very helpful for this, allowing something to act like it's in the mod.rs file for visibility purposes but not having to actually be in the mod.rs) Examples of this are in asset_manager.rs, resource_manager.rs, and descriptor_set_allocator.rs. They are the "top-level" interfaces for the "folder" they are in. The mod.rs in those cases contains small misc. bits or the using statements. I think 1, 2, and 3 are fine to be in the top level mod file. I think 3 is fine to be stored along with 4 (as it is now) because in *most* cases it's not a public-facing struct. However, it's likely to be a small type so I don't have a strong opinion on it.

demo/src/features/debug3d/mod.rs
aclysma: Looking at the features we have now, only 1/6 of them have more than one module that we would re-export publicly. So I think we will end up with a lot of public folders containing an almost empty mod.rs and a single .rs file with the public-facing type.

If anything, I would suggest have a jobs folder and moving the extract/prepare/write into that and leaving the other public modules in the root. I don't mind an "internal" module/folder as much but I think it too will rarely contain more than 1 file. In general, I prefer a flat structure unless there are >3 files that can be grouped into a directory.

Would having a jobs directory and leaving the public modules in the root feel more "readable" to you? Then you could do pub use crate::feature_name::* and get all the public symbols.

demo/src/features/skybox/extract.rs
aclysma: Been thinking about this today:
 * I still think that the 10x increase in amount of code is not worth the consistency gained. (Also that it's "negative teaching" because this isn't the way I would recommend someone to add a full-screen triangle pass.)
 * A skybox seems to match the green text rather than the yellow text. I doubt they went through their nodes system for their "atmosphere" pass.
 * The current form of the skybox is ideally drawn after `OpaqueRenderPhase` and before `TransparentRenderPhase`. (Although the performance difference is likely not measurable on modern hardware.)
 * What I would like to do longer term is add an obvious "correct" way to do a full-screen pass. We do this in 4 places in the current render graph (see cmd_draw(3, 0)) vs 5 call sites for `write_view_phase`.

However, since I don't think the current implementation is ideal, if you feel strongly that doing it as a feature is better, we can do it that way now. I would like to find a better way to support inserting a "full screen pass" later that is less code than either of these approaches because it seems like a common operation.

demo/src/features/debug3d/mod.rs
DavidVonDerau: So the proposed directory would be something like:

```
.\feature_name\
    .\internal\
        .\<optional helper code>.rs
    .\jobs\
        extract.rs
        prepare.rs
        write.rs
    .\mod.rs
    .\plugin.rs
    .\<optional top level interfaces>.rs
```

`internal` wouldn't be present unless it was actually needed. `jobs` would always exist. There would be no more `public` directory, because anything at the top-level _is_ public.

I can be fine with that.

demo/src/features/debug3d/mod.rs
DavidVonDerau: See reply here. https://github.com/aclysma/rafx/pull/122#discussion_r610996877

demo/src/features/debug3d/write.rs
DavidVonDerau: For the mesh and tile layer, I agree, those should be locally in a clearly public location. The `tile_layer` `TileLayerVertex` is currently in the `public` directory, but pending https://github.com/aclysma/rafx/pull/122#discussion_r610996877 I'll move it into the top-level folder.

I'm less convinced on `Debug3DVertex` because it's not used outside of the feature, and I'm not sure in what scenario it would be.

demo/src/features/debug3d/extract.rs
aclysma: I tried to find some alternatives.
 * #[profiling::function] makes a massive scope name (`<demo::features::debug3d::extract::Debug3dExtractJob as rafx_framework::nodes::jobs::extract::ExtractJob>::extract`)
 * scope!(std::any::type_name::<self>()) is still long (`demo::features::debug3d::extract::Debug3dExtractJob`)
 * I also tried calling `self.feature_debug_name()` but couldn't find a good way to concat that at compile time

Using super::extract_scope - using super:: does help a lot with knowing where this actually is located. I still don't like it being generated by a macro (feels like magic in a bad way). But if you feel strongly about this over just putting in a string, I'm willing to go ahead with it. But lets name it EXTRACT_SCOPE_NAME.

demo/src/features/debug3d/mod.rs
aclysma: yeah, I'm find with this too!

demo/src/features/debug3d/write.rs
aclysma: * if it's public I think we agree the vertex declaration can be in the top-level mod.rs (it's very small, doesn't justify its own .rs file)
 * I think we also agree it's best to have it in a consistent place
 * So I personally think we should always put it in the top-level mod.rs so it's in the same place for all features and just add pub as needed

demo/src/features/debug3d/extract.rs
DavidVonDerau: Done.

demo/src/features/debug3d/write.rs
DavidVonDerau: Now that I've done a restructure of everything, I kind of like the super minimal `mod.rs` file. It's _just_ the feature declaration, private `use` includes, and the public API reexport.

If the goal is to have it all in a consistent place, I would honestly prefer moving it to the `write.rs` file and just re-exporting it as public in the top-level `mod.rs` when needed (e.g. for `MeshVertex` or `TileLayerVertex`). I think that's nicer because it's next to the vertex declaration too.

DavidVonDerau: e.g.

```rust
mod jobs;
use jobs::*;

use rafx::render_feature_mod_prelude::*;
rafx::declare_render_feature!(TileLayerRenderFeature, TILE_LAYER_FEATURE_INDEX);

// Public API

mod plugin;
pub use plugin::*;

mod tile_layer_render_node_set;
pub use tile_layer_render_node_set::*;

mod tile_layer_resource;
pub use tile_layer_resource::*;

pub use jobs::TileLayerVertex;
```

demo/src/features/mesh/shadow_map_resource.rs
```diff
@@ -28,15 +43,15 @@ enum ShadowMapVisibility {
 #[derive(Default)]
 pub struct ShadowMapResource {
     // These are populated by recalculate_shadow_map_views()
-    pub(super) shadow_map_lookup: FnvHashMap<LightId, usize>,
-    pub(super) shadow_map_render_views: Vec<ShadowMapRenderView>,
+    pub(in crate::features::mesh) shadow_map_lookup: FnvHashMap<LightId, usize>,
```
aclysma, line 46: pub(super) doesn't work here?

demo/src/features/text/text_resource.rs
```diff
@@ -89,7 +67,7 @@ impl TextResource {
     }
 
     // Returns the draw data, leaving this object in an empty state
-    pub fn take_text_draw_data(&mut self) -> TextDrawData {
+    pub(in super::super) fn take_text_draw_data(&mut self) -> TextDrawData {
```
aclysma, line 70: maybe this should just be pub

demo/src/features/debug3d/jobs/prepare.rs
```diff
@@ -1,51 +1,55 @@
...
+
         let per_view_descriptor_set_layout =
             &self.debug3d_material_pass.get_raw().descriptor_set_layouts
                 [shaders::debug_vert::PER_FRAME_DATA_DESCRIPTOR_SET_INDEX];
 
-        let mut per_view_descriptor_sets = Vec::default();
-        for view in views {
+        for view in views
+            .iter()
+            .filter(|view| view.feature_is_relevant::<RenderFeatureType>())
```
aclysma, line 51: I kind of dislike having this typedef. It makes it impossible to search the code for all occurrences of a particular feature. I do realize this is something that has to be changed when copy-pasting an existing feature to make a new one, but the person doing it would already need to search-replace OldFeatureName with NewFeatureName and old_feature_name with new_feature_name so I don't see it as a big deal.

aclysma (commented):
A few more minor comments

demo/src/features/debug3d/jobs/prepare.rs
aclysma: (this comment is meant for all usages of RenderFeatureType as a typedef, not just this particular example)

demo/src/features/mesh/shadow_map_resource.rs
DavidVonDerau: Fixed.

demo/src/features/text/text_resource.rs
DavidVonDerau: Replaced `in super::super` with just `super`.

demo/src/features/debug3d/jobs/prepare.rs
DavidVonDerau: Really? I felt like it made it _easier_ to search the code.

With the typedef, when you search for a feature, you'll find exactly the following:
1. The definition of it inside of the feature `mod.rs` file.
2. Any external uses of it by importing or reference.

Without the typedef, when you search for a feature, you'll find the following:
1. The definition of it inside of the feature `mod.rs` file.
2. Any external uses of it by importing or reference.
3. An arbitrary number of internal uses of the feature type inside of the feature directory, drowning out everything else.

e.g. compare `MeshRenderFeature` 

image

vs

image

aclysma: Yeah, the typedef makes it so searching for MeshRenderFeature omits the internal usages which is misleading. If it’s used in 15 places, I do want to see all those 15 places. :)

DavidVonDerau: :( Very well, I'll nuke the typedef later today.

PR #125, opened 2021-04-15
Add rafx-visibility & frustum culling to rafx-framework, renderer, demo

DavidVonDerau:
`rafx-visibility` is a new low-level crate used by `rafx-framework` for providing a `VisibilityWorld`. A `VisibilityWorld` is like a physics world -- it's a retained API that maintains a shadow state of the application's game world. The API encapsulates the game code from changes to the visibility's code algorithms or data structures. As new features are added to `rafx-visibility` like occlusion culling or hierarchical cull structures for large game scenes, those features will become available transparently to the applications relying on `rafx-visibility`.

### Public API

The public API to `rafx-visibility` is contained in `src/*` and `src/geometry/*`.

The main entry point is the `VisibilityWorldArc`. The application creates `VisibilityWorldArc` and registers new `Objects`, `Models`, `Zones`, `Volumes`, and `View Frustums` with it to mirror their game's scene. Additionally, the application can use a channel to send asynchronous commands to the `VisibilityWorldArc` in a thread-safe manner that can be processed during the next update of the `VisibilityWorld`.

The `VisibilityWorld` uses handles to avoid bleeding internal design into the application.

- A `ViewFrustumHandle` represents a view in the application, like the main camera, a mini-map, or a light source needed for shadow casting. It has a specific `Projection`. The supported `Projection` types are `Orthogonal` or `Perspective`.
- An `ObjectHandle` represents a game entity in the application. It is associated with a `Model` for culling and placed into a `Zone` with `View Frustums`.
- A `ZoneHandle` is used to group `Objects` and `View Frustums` together. `View Frustums` are only able to "see" `Objects` in the same `Zone`. `Objects` and `View Frustums` may not be in multiple zones simultaneously. This is similar to a layer in a physics API. In a future release, the intent is for `Zones` to be the way to connect easily segmented levels with `Portals`. [1]
- A `ModelHandle` represents the visibility bounds for a game entity in the application. Internally, it's represented by one of several possible bounding structures. In a physics API, this is like defining the collider for an entity in the physics world.
- A `VolumeHandle` is like a `Model`, but instead of being reduced down to the bounding structure, it maintains a well-defined geometric shape like a cone, sphere, capsule, or other supported 3-dimensional shape. Unlike a `Model`, a `Volume` is not associated with an `ObjectHandle`. It is placed into a `Zone` and positioned independently. `Volumes` are included in the list of visibility results from a `View Frustum` when the `Volume` intersects with that `View Frustum`. `Volumes` are currently not implemented. In the future, `Volumes` will be the primary way for the application to determine the visibility of light sources in a view for the purpose of culling unseen lights. [2]

The public API also includes a `VisibleBounds` struct for processing an arbitrary `PolygonSoup` provided by the application into the `VisibilityWorld`'s internal bounding structures. `VisibilityBounds` may be computed offline and provided to the `VisibilityWorld` during loading for initializing `Models`. The reason for preferring an internal structure is that the `VisibilityWorld` will need to simplify application models into simpler `OccluderBounds` in the future when occlusion culling is added.

### Visibility Queries

The `VisibilityWorldArc` defines the `query_visibility` function. This function is thread-safe.

```rust
pub fn query_visibility(
    &self,
    view_frustum: ViewFrustumHandle,
    result: &mut VisibilityQuery,
) -> Result<(), QueryError>
```

The result of a `VisibilityQuery` is a `Vec` of `VisibilityResult<T>` where `T` is either an `ObjectHandle` or a `VolumeHandle`. Each `VisibilityResult` contains the following information:

```rust
pub struct VisibilityResult<T> {
    pub handle: T,
    pub id: u64,
    pub bounding_sphere: BoundingSphere,
    pub distance_from_view_frustum: f32,
}
```

- `handle` is the `ObjectHandle` or `VolumeHandle` visible to the `ViewFrustumHandle` during the query.
- The `id` is an application-provided value set & retained on each `ObjectHandle` or `VolumeHandle` with the `set_object_id` and `set_volume_id` functions. The `id` is intended to be used by the application as a key back to their game entity, e.g. a `ptr`, or an `Entity ID` in an `ECS`, or a key in some type of `Map`.
- The `bounding_sphere` and `distance_from_view_frustum` are provided to support applications maintaining a level-of-detail budget. [3]

## Internals

- `src/internal/*` defines the data structures and algorithm for thread-safe frustum culling.
- `src/frustum_culling/*` contains an SIMD algorithm for culling bounded spheres quickly and associated data structure -- the `PackedBoundingSphere` and `PackedBoundingSphereChunk`. [4]

[1] https://en.wikipedia.org/wiki/Portal_rendering

[2] Lengyel, Eric. (2019). Foundations of Game Engine Development: Vol 2. Rendering.

[3] http://advances.realtimerendering.com/destiny/gdc_2015/Tatarchuk_GDC_2015__Destiny_Renderer_web.pdf

[4] https://www.ea.com/frostbite/news/culling-the-battlefield-data-oriented-design-in-practice


demo/assets/ldtk/example.ldtk.meta
```diff
@@ -56,4 +56,4 @@
         ],
         ldtk_asset_uuid: Some("8273ae0f-73fc-4037-bf83-a34f6e3b6c13"),
     ),
-)
```
DavidVonDerau, line 59: Spurious change

demo/src/features/mesh/shadow_map_resource.rs
```diff
@@ -275,27 +181,41 @@ fn calculate_shadow_map_views(
...
-        let light_to = position.position + light.direction;
+        let eye_position = transform.translation;
+        let light_to = transform.translation + light.direction;
 
         let view = glam::Mat4::look_at_rh(eye_position, light_to, glam::Vec3::new(0.0, 0.0, 1.0));
 
-        let near_plane = 0.25;
+        let near_plane = 0.1;
         let far_plane = 100.0;
         let proj = perspective_rh(light.spotlight_half_angle * 2.0, 1.0, far_plane, near_plane);
```
DavidVonDerau, line 194: Use `as_mat4` function on `Projection`.

demo/src/features/mesh/shadow_map_resource.rs
```diff
@@ -313,7 +233,7 @@ fn calculate_shadow_map_views(
             glam::Vec3::new(0.0, 0.0, 1.0),
         );
 
-        let near_plane = 0.25;
+        let near_plane = 0.1;
         let far_plane = 100.0;
         let ortho_projection_size = 10.0;
         let proj = glam::Mat4::orthographic_rh(
```
DavidVonDerau, line 239: Use `as_mat4` function on `Projection`.

demo/src/features/mesh/shadow_map_resource.rs
```diff
@@ -370,30 +307,44 @@ fn calculate_shadow_map_views(
                 cube_map_view_directions.1,
             );
 
-            let near = 0.25;
+            let near = 0.1;
             let far = light.range;
             let proj = glam::Mat4::perspective_lh(std::f32::consts::FRAC_PI_2, 1.0, far, near);
```
DavidVonDerau, line 312: Use `as_mat4` function on `Projection`.

demo/src/features/tile_layer/tile_layer_resource.rs
```diff
@@ -50,13 +51,21 @@ impl TileLayerResource {
...
+                    let handle = visibility_region.register_static_object(
+                        EntityId::from(tile_layer_entity_id),
+                        CullModel::quad(layer.width as f32, layer.height as f32),
+                    );
+                    let mut translation = layer.center;
+                    translation.y = -translation.y; // NOTE(dvd): +y is up in our world, but _down_ in LDtk.
+                    handle.set_transform(translation, Quat::IDENTITY, Vec3::ONE);
+                    handle.add_feature(render_node.as_raw_generic_handle());
+
+                    log::info!("{} {} {:?}", layer.width, layer.height, layer.center);
```
DavidVonDerau, line 65: Remove debug message.

demo/src/scenes/mod.rs
```diff
@@ -159,6 +159,13 @@ fn add_light_debug_draw(
 ) {
     let mut debug_draw = resources.get_mut::<DebugDraw3DResource>().unwrap();
 
+    /*
```
DavidVonDerau, line 162: Remove commented out code.

rafx-visibility/src/camera.rs
```diff
@@ -0,0 +1,409 @@
...
+trait UpdateFrustum {
+    fn update_frustum(
+        &self,
+        camera: &Camera,
+        frustum: &mut Frustum,
+    );
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Projection {
```
DavidVonDerau, line 182: Projection code can probably be in a separate file.

rafx-visibility/src/geometry/transform.rs
```diff
@@ -0,0 +1,36 @@
+use glam::{Mat4, Quat, Vec3};
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct Transform {
```
DavidVonDerau, line 4: This should probably be private to the crate if it isn't already.

demo/src/features/mesh/shadow_map_resource.rs
```diff
@@ -370,30 +307,44 @@ fn calculate_shadow_map_views(
                 cube_map_view_directions.1,
             );
 
-            let near = 0.25;
+            let near = 0.1;
```
aclysma, line 310: You mentioned previous some visual artifacts, I've noticed that changing the near field can have a strong affect on precision in the depth buffer (even when we use the reverse z trick). It's *possible* if you still notice them, that this may be related. (I wouldn't expect it to be a problem though.)

rafx-framework/Cargo.toml
```diff
@@ -27,6 +29,8 @@ profiling = "0.1.9"
 serde = { version = "1", features = ["serde_derive"] }
 serde_bytes = "0.11"
 glam = { version = "0.13.1", features = [ "serde" ] }
+parking_lot = "0.11.1"
+slotmap = "1.0.2"
```
aclysma, line 33: I want to look into slotmap a bit more later, seems similar to the GenSlab this PR removes (but maybe it does more stuff or implements things in a better way than GenSlab)

rafx-framework/Cargo.toml
```diff
@@ -27,6 +29,8 @@ profiling = "0.1.9"
 serde = { version = "1", features = ["serde_derive"] }
 serde_bytes = "0.11"
 glam = { version = "0.13.1", features = [ "serde" ] }
+parking_lot = "0.11.1"
```
aclysma, line 32: another note to myself to come back and look at exactly why parking_lot is needed, if we already pull it in as a dependency, or pull in an alternative to it that could fill the same role.

rafx-framework/src/visibility/camera_object.rs
```diff
@@ -0,0 +1,221 @@
...
+use parking_lot::{RwLock, RwLockReadGuard};
+use rafx_api::RafxError;
+use rafx_visibility::{
+    AsyncCommand, CameraHandle, Projection, RegionHandle, VisibilityQuery, VisibilityWorldArc,
+};
+use slotmap::SlotMap;
+use std::sync::Arc;
+
+#[derive(Clone)]
+pub struct CameraObjectHandle {
```
aclysma, line 14: I think we should consider some other names instead of camera. I would expect a camera to be a more engine-level thing that would include more tangible, artistic/PBR parameters that drive rendering (exposure, lens settings, dof, etc.). This is a fairly abstract system. In other words, I would expect end-user code (or even we create a separate high level system) to create its own thing called a Camera (likely bindable to render to a viewport on a window) and it would own this more abstract "view/query" object. (Similarly to how lights own these objects.) Maybe something with the word Frustum in it?

rafx-visibility/src/frustum_culling/packed_bounding_sphere_chunk.rs
```diff
@@ -0,0 +1,367 @@
...
+            }
+        } else {
+            BoundingSphere {
+                position: Vec3::new(self.x2, self.y2, self.z2),
+                radius: self.radius2,
+            }
+        };
+    }
+
+    #[inline(always)]
```
aclysma, line 313: did you measure this to be an improvement?

aclysma: (the `#[inline(always)]` I mean)

aclysma:
Great stuff! The main two things are
1) I want to look into the added dependencies a little
2) Decide on if we should use the word camera or something else. Even if we use the word camera, at least we could add the word visibility (like VisibilityCamera) to make it clear that it's part of the visibility system, not a scene-level thing that drives rendering to a viewport on the screen.

rafx-visibility/src/frustum_culling/packed_bounding_sphere_chunk.rs
DavidVonDerau: Yes, at least on my machine. 🤷

demo/src/features/mesh/jobs/prepare.rs
```diff
@@ -481,6 +481,8 @@ impl PrepareJob for MeshPrepareJob {
             for (frame_node_index, frame_node_data) in
                 self.extracted_frame_node_mesh_data.iter().enumerate()
             {
+                // TODO(dvd): we should only prepare render nodes that are different
```
DavidVonDerau, line 484: Remove this comment.

rafx-visibility/src/internal/visible_model.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[derive(Default, Copy, Clone)]
+pub struct VisibleModel {
+    pub aabb: AxisAlignedBoundingBox,
+    pub obb: OrientedBoundingBox,
+    pub bounding_sphere: BoundingSphere,
+    pub hash: u64,
+}
+
+// TODO(dvd): Extract this code so that it can be used in the importer.
```
DavidVonDerau, line 13: Implement this before merge.

rafx/src/lib.rs
```diff
@@ -85,5 +85,8 @@ pub mod render_feature_mod_prelude {
     pub use rafx_framework::nodes::{RenderFeature, RenderFeatureIndex};
     pub use std::convert::TryInto;
 }
+#[cfg(feature = "framework")]
```
DavidVonDerau, line 88: Missing a newline?

rafx-framework/src/visibility/camera_object.rs
DavidVonDerau: Will rename to `VisibilityFrustum`

README.md
```diff
@@ -159,6 +161,9 @@ This tool currently is only useful for packing assets.
  * `rafx-api` - Rendering API abstraction layer.
    * Vulkan backend for windows/linux
    * Metal backend for macOS/iOS
+ * `rafx-visibility` - Retained visibility API.
```
aclysma, line 164: Honestly I'm not sure it makes a lot of sense to split the visibility code across two crates. Rafx started out with more crates (including a Rafx-visibility crate), but I ultimately decided it was not helpful and added more maintenance/complexity. (Nothing major or anything, but it was a constant cost that benefited no one.) I also think it's more friendly and approachable to users to understand the crates as a linear progression - i.e. a layer cake rather than a tree.

It's true that separating crates is *technically* more flexible, but I'm not sure who would actually use rafx-visibility but not rafx-framework. I'm willing to try it the way you have it now, but I wanted to mention it because I did already try it before and the experience made me want to have fewer crates.

README.md
```diff
@@ -171,6 +176,8 @@ This tool currently is only useful for packing assets.
         actually executed parallel yet)
       * Handles multiple views and phases allowing advanced features like shadow mapping
       * Flexible sorting mechanism for interleaving and batching write commands from multiple rendering features
+   * Visibility Region - Built on top of `rafx-visibility`
+      * Provides reference counted wrappers around the low-level handles and a more ergonomic interface that assumes a clear definition of `static` and `dynamic` game geometry.
```
aclysma, line 180: I didn't really understand what "assumes a clear definition of `static" means. I think this might also be too much detail. "Higher-level visibility API, integrated with framework render jobs" or something like that I think is enough.

docs/visibility/api_design.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Public API
+
+The public API to `rafx-visibility` is contained in `src/*` and `src/geometry/*`.
+
+The main entry point is the `VisibilityWorldArc`. The application creates `VisibilityWorldArc` and registers new `Objects`, `Models`, `Zones`, `Volumes`, and `View Frustums` with it to mirror their game's scene. Additionally, the application can use a channel to send asynchronous commands to the `VisibilityWorldArc` in a thread-safe manner that can be processed during the next update of the `VisibilityWorld`.
+
+The `VisibilityWorld` uses handles to avoid bleeding internal design into the application.
+
+- A `ViewFrustumHandle` represents a view in the application, like the main camera, a mini-map, or a light source needed for shadow casting. It has a specific `Projection`. The supported `Projection` types are `Orthogonal` or `Perspective`.
```
aclysma, line 15: At some point we should figure out a consistent naming style for suffixes like Handle. Should the suffixes "Handle" and "Arc" have some consistent semantic meaning? (i.e. handle has to be deleted manually, arc can simply be dropped.) This is probably a larger conversation to have about rafx as a whole, not just this PR. Happy to merge this now and revisit later.

I'm also wondering if we should prefix these types with something like "Viz". (Kind of how all the low-level API types are prefixed "Rafx", I think it can help with readability.) The main concern is that sometimes names get kind of generic (like "ModelHandle" and "ZoneHandle" aren't obvious that they are visibility-related, and it's very plausible someone could have their own concept of Zone or Model in their code that uses rafx.)

rafx-framework/Cargo.toml
aclysma, line 32: wondering if we should consistently use parking_lot::Mutex everywhere. (And maybe typedef a RafxMutex so that this is decided upon in one place.)

rafx-framework/src/visibility/entity_id.rs
```diff
@@ -0,0 +1,25 @@
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct EntityId(u64);
```
aclysma, line 2: I think this is a case where, if we had a prefix on names, it would be helpful to end-users because they wouldn't have to deal with having two EntityId structs simultaneously in their code. I don't have a specific suggestion for that though and it's probably a question that's outside the scope of this PR.

rafx-framework/src/visibility/view_frustum_object.rs
```diff
@@ -0,0 +1,239 @@
...
+        if let Some(dynamic_view_frustum) =
+            self.view_frustum(&self.inner.dynamic_view_frustum, &storage)
+        {
+            dynamic_view_frustum.set_projection(projection);
+        }
+
+        self
+    }
+
+    pub fn query_visibility(&mut self) -> RafxResult<RwLockReadGuard<VisibilityQuery>> {
```
aclysma, line 103: returning a lock feels like a deadlock waiting to happen (but I don't have a useful alternative to suggest atm :D )

rafx-framework/src/nodes/frame_packet.rs
```diff
@@ -179,22 +185,33 @@ impl FramePacketBuilder {
     pub fn add_view(
```
aclysma, line 185: we may want to reconsider the name for this function since now it's doing the full visibility check, (assuming I understand the code correctly.) Or make this function take results and leave it to the user to call the query function. This would make it more clear to a user that it's potentially a very expensive operation.

rafx-framework/src/visibility/visibility_region.rs
```diff
@@ -0,0 +1,104 @@
...
+    static_zone: ZoneHandle,
+}
+
+impl VisibilityRegion {
+    pub fn new() -> Self {
+        let visibility_world = VisibilityWorldArc::new();
+        let allocator = Arc::new(VisibilityObjectAllocator::new(visibility_world.clone()));
+
+        let static_zone = {
+            let mut inner = visibility_world.inner.lock();
```
aclysma, line 24: very minor thing but I imagine we could only lock once here, or not lock at all if we defer putting visibility_world into an Arc until after we finish populating it

rafx-visibility/src/internal/visibility_world.rs
```diff
@@ -0,0 +1,568 @@
...
+        self.objects.remove(object).unwrap();
+    }
+
+    // --------
+    // Volumes
+    // --------
+
+    /// Creates a new `Volume`. An `Volume` must be in a `Zone` to be visible.
+    pub fn new_volume(&self) -> VolumeHandle {
+        unimplemented!();
```
aclysma, line 409: I would lean towards commenting these out

rafx-framework/src/nodes/frame_packet.rs
aclysma: query_visibility_and_add_results is wordy but would make it clear what this function is doing and that it's a significant operation

DavidVonDerau: I'd like to punt on changing the name because I imagine that this function will need to be modified very soon -- once we expose callbacks for the user to do their own threading, we'll have to yank the visibility code out of here. That was something I was going to look at as part of this work, but as the lines of code kept piling up I felt it was more important to ship a first increment and then iterate on it.

rafx-framework/src/visibility/entity_id.rs
DavidVonDerau: `OpaqueId`? `CallbackId`?

rafx-framework/Cargo.toml
DavidVonDerau: The parking_lot::Mutex is a bit of a double-edged sword if you wanted to typedef it. The problem is that it's more ergonomic to use _but_ that also means it's not "just" a drop-in replacement. You don't need a separate line to declare a guard and you don't need an extra `unwrap`. 

So if we typedef'd it, we'd still need to change every line of code using it when the type under the typedef changed.

docs/visibility/api_design.md
DavidVonDerau: Hmm, I kind of like the idea of using `Handle` to mean "this must be deleted manually" vs `Arc` implies "you can just drop it". I'd have to go & update a bunch of the names but I don't mind.

DavidVonDerau: > I'm also wondering if we should prefix these types with something like "Viz". (Kind of how all the low-level API types are prefixed "Rafx", I think it can help with readability.) The main concern is that sometimes names get kind of generic (like "ModelHandle" and "ZoneHandle" aren't obvious that they are visibility-related, and it's very plausible someone could have their own concept of Zone or Model in their code that uses rafx.)

Isn't that what namespaces are intended to solve? In the event that scenario occurred, they could refer to the `Rafx` types using a partially or fully qualified name.

rafx-framework/src/visibility/view_frustum_object.rs
DavidVonDerau: Yeah, I'm not happy about it either. We'll be able to find a better API as we use it more in examples and continue bringing in functionality.

docs/visibility/api_design.md
aclysma: We could do renaming in a follow-up PR if we decide on a specific plan. This is already a big PR and will get unwieldy if it sits long. (We can add issues for anything we decide to do later)

I still try to avoid name collisions because I think it makes code unnecessarily hard to read when some places use a::TypeName and other places use b::TypeName.. every time you see TypeName somewhere you have to look at what was imported and it's likely not to be consistent across files. Even if it's always fully specified, it feels a lot noisier to have to namespace everything. I also think it's valuable in this case that anyone reading code would know all the things beginning with "Viz" (or some other prefix) are closely related.

rafx-framework/Cargo.toml
aclysma: Yeah I don't think we need to make it swappable with no code changes. It would mainly be to make it clear that this is the "correct" mutex to use in rafx code. We don't need to decide on this now, just a thought.

rafx-framework/src/visibility/entity_id.rs
aclysma: I lean towards either UserId or just stick with EntityId. None of them feel vastly better than the others to me.

rafx-framework/src/nodes/frame_packet.rs
aclysma: We could temporarily rename it to query_visibility_and_add_results for now (since it's an easy/quick change) and let that motivate us to find something better later.

aclysma: (I do agree with the sentiment that we shouldn't pile on more things to an already-large PR, and this is a large enough feature that it will need to be iterated on)

README.md
aclysma: Even if we agree to merge the crates, I'm happy for us to put this in as it is and change it later.

rafx-framework/src/visibility/entity_id.rs
DavidVonDerau: Renamed to `UserId`.

README.md
DavidVonDerau: I have three at least three reasons why I'd like to keep it separate.

1. I feel somewhat more confident in the low-level API than I do in something higher-level with ref-counted smart pointers. I already ran into a couple of thorny problems implementing the `VisibilityRegion`, so keeping this separated is a way of insulating each side from the churn of the other side (assuming it's not along that shared boundary API)
2. I can foresee the lower-level crate getting progressively more complex (in terms of lines of code) to handle things like occlusion culling or hierarchical culling, whereas I wouldn't expect that to occur on the higher-level code since it's mostly providing the convenience of a more ergonomic interface, vs. functionality.
3. Where the higher-level code _does_ introduce new functionality, it's unrelated to the core problem of visibility -- e.g. associating visibility objects with render features.

README.md
```diff
@@ -8,6 +8,7 @@ for real-world projects with multidisciplinary teams.
 
 This crate contains several layers:
  * `rafx-api`: Low-level graphics API abstraction 
+ * `rafx-visibility`: Low-level visibility API abstraction
```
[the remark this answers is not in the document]
DavidVonDerau: Changing to `Low-level retained API for visibility determination`.

docs/visibility/api_design.md
DavidVonDerau: re: offline discussion. I'll rename the higher-level visibility structs since it's awkward having them both be called `Handle`.

README.md
aclysma: We could keep the two as separate modules within a single crate. But I'm fine either way. This is not hard to switch if we change our mind later :)

README.md
```diff
@@ -24,7 +25,7 @@ Please keep in mind, this crate is still in pre-0.1.0 status!
 
 [![Build Status](https://github.com/aclysma/rafx/workflows/CI/badge.svg)](https://github.com/aclysma/rafx/actions)
 
-[![Video of Renderer in Use](docs/ios-screenshot.png)](https://www.youtube.com/watch?v=Ks_HQbejHE4 "Video of Renderer in Use")
```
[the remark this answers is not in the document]
DavidVonDerau: Copied & reverted that line.

Issue #126, opened 2021-04-20
rafx_framework with winit on macos panics (unwrap() on crossbeam SendError) when droping resources.


jlowry:
Here is some code that causes the problem: https://github.com/jlowry/laughing-waffle/blob/81201f87c5e840b86e92796fa59245b67e241e96/src/main.rs

```
[2021-04-20T14:25:24.338132000Z INFO  rafx_framework::resources::resource_manager] Cleaning up resource manager
[2021-04-20T14:25:24.338640000Z WARN  rafx_framework::resources::resource_lookup] rafx_framework::resources::resource_lookup::MaterialPassResource resource count 1 > 0, resources will leak
[2021-04-20T14:25:24.338695000Z WARN  rafx_framework::resources::resource_lookup] rafx_framework::resources::resource_lookup::DescriptorSetLayoutResource resource count 1 > 0, resources will leak
[2021-04-20T14:25:24.338716000Z WARN  rafx_framework::resources::resource_lookup] rafx_framework::resources::resource_lookup::RootSignatureResource resource count 1 > 0, resources will leak
[2021-04-20T14:25:24.338739000Z WARN  rafx_framework::resources::resource_lookup] rafx_framework::resources::resource_lookup::ShaderResource resource count 1 > 0, resources will leak
[2021-04-20T14:25:24.338758000Z WARN  rafx_framework::resources::resource_lookup] rafx_framework::resources::resource_lookup::ShaderModuleResource resource count 2 > 0, resources will leak
[2021-04-20T14:25:24.339160000Z INFO  rafx_framework::resources::resource_manager] Dropping resource manager
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', /Users/james/.cargo/git/checkouts/rafx-e743dbb61b0ed536/515aeea/rafx-framework/src/resources/resource_arc.rs:53:50`
```

jlowry:
I'd be interested in a having a go at fixing this.

Any thoughts on how it should be fixed? I'm assuming that it's not a simple as handling the SendError instead of unwrap. Do we need to make sure the resource manger is not dropped until after the assets?

aclysma:
Yeah, things need to be destroyed in the opposite order of creation. I am using rafx with winit/macOS so I can confirm this combination works.

It would probably help to make that unwrap and expect with a more helpful error message, I can make that change later.

jlowry:
[...]

PR #132, opened 2021-05-02
Change destroying resource after its manager unwrap to expect

aclysma:
@DavidVonDerau have any thoughts on this? I think we could either make it a log (non-fatal, but likely happening when the program is shutting down anyways) or an expect() which would be fatal but still give more clarity than a simple unwrap. There is already helpful warning spam when the resource manager gets shut down early. I'm usually happy to fail fast and loud but given this is happening when the program is likely already shutting down, I don't mind relaxing it a bit.

This is based on #126

### DIFF
--- a/rafx-framework/src/resources/resource_arc.rs
+++ b/rafx-framework/src/resources/resource_arc.rs
@@ -50,9 +50,8 @@ where
     ResourceT: Clone,
 {
     fn drop(&mut self) {
-        if self.drop_tx.send(self.resource.clone()).is_err() {
-            log::warn!("A resource was dropped after the resource manager was destroyed, it might not clean up properly")
-        }
+        self.drop_tx.send(self.resource.clone())
+            .expect("A resource was dropped after the resource manager was destroyed, it might not clean up properly");
     }
 }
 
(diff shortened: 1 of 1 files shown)


rafx-framework/src/resources/resource_arc.rs
```diff
@@ -50,7 +50,9 @@ where
     ResourceT: Clone,
 {
     fn drop(&mut self) {
-        self.drop_tx.send(self.resource.clone()).unwrap();
+        if self.drop_tx.send(self.resource.clone()).is_err() {
+            log::warn!("A resource was dropped after the resource manager was destroyed, it might not clean up properly")
```
DavidVonDerau, line 54: I wonder if we should have an optional feature for "validation-mode". When validation mode is on, this is an expect / panic. When it's off, it's a warn statement. I imagine there are many other places in the code where we'd like to fail fast and loudly during development but relax it for release.

aclysma: made it an expect

DavidVonDerau (approved):

DavidVonDerau:
Nit: PR title needs to be updated.

PR #133, opened 2021-05-08
Ability to pause visibility updates, some other small fixes

aclysma:
- Debug mode to disable visibility updates 
- Fix suspected vulkan sync issue in swapchain handling
- Reorder main update loop to ensure anything that uses the window size occurs after the event loop (fixes imgui bug)
- Fix vulkan not initializing validation layers


rafx-framework/src/visibility/view_frustum_arc.rs
```diff
@@ -100,13 +101,18 @@ impl ViewFrustumArc {
...
+        &mut self,
+        visibility_config: &VisibilityConfig,
+    ) -> RafxResult<RwLockReadGuard<VisibilityQuery>> {
         let mut results = self.inner.visibility_query.write();
 
         results.objects.clear();
         results.volumes.clear();
 
-        self.inner.visibility_world.update();
+        if visibility_config.enable_visibility_update {
```
DavidVonDerau, line 113: I think I would probably use this to skip the `clear` on the 2 lines above, and then skip lines 119 through 129 (the actual calls to `query_visibility`). The issue with skipping the updates to the visibility world is that I think you'll be piling up channel updates (a memory leak, of sorts).

demo/src/lib.rs
```diff
@@ -241,6 +245,11 @@ impl RenderOptions {
             combo.end(ui);
             self.tonemapper_type = current_tonemapper_type.into();
         }
+
+        ui.checkbox(
+            imgui::im_str!("freeze visibility"),
+            &mut self.freeze_visibility,
```
DavidVonDerau, line 251: Would probably rename this to `enable_visibility_update` and default it to `true`
That would keep consistency with other settings in this menu (checked is active, unchecked is off) + it means that you wouldn't need to negate it when you copy it into `VisibilityConfig`.

DavidVonDerau (approved):
Approved, assuming the CI passes.

demo/src/lib.rs
```diff
@@ -344,6 +342,23 @@ pub fn run(args: &DemoArgs) -> RafxResult<()> {
...
+        //
+        // Process input
+        //
+        if !process_input(&mut scene_manager, &mut world, &resources, &mut event_pump) {
+            break 'running;
+        }
+
+        {
+            let mut viewports_resource = resources.get_mut::<ViewportsResource>().unwrap();
+            let (width, height) = sdl2_systems.window.vulkan_drawable_size();
```
DavidVonDerau, line 354: @aclysma Should this be `size()` instead of `vulkan_drawable_size()`?

aclysma: I'm genuinely not sure. I made this change because the imgui code was asking for a viewport size bigger than the window. I thought there was a mismatch between size() and vulkan_drawable_size(). However, the root cause was that I was ticking imgui before sdl event processing, and then doing rendering. The size changed when processing sdl events. So imgui's window size and the swapchain size did not match.

I think the difference between the two is how hidpi monitors are handled. To some degree, that might be an opinion more than a "right/wrong" thing.

PR #135, opened 2021-05-12
Replace `nodes` with `render_features` and new jobs framework.

DavidVonDerau:
If you are looking for an example of the new entry points, look at the `Mesh` feature in the demo.
`RenderFeaturePlugin`: https://github.com/DavidVonDerau/rafx/blob/dvd/jobs/demo/src/features/mesh/plugin.rs
`RenderFeatureExtractJob`: https://github.com/DavidVonDerau/rafx/blob/dvd/jobs/demo/src/features/mesh/jobs/extract.rs
`RenderFeaturePrepareJob`: https://github.com/DavidVonDerau/rafx/blob/dvd/jobs/demo/src/features/mesh/jobs/prepare.rs
`RenderFeatureWriteJob`: https://github.com/DavidVonDerau/rafx/blob/dvd/jobs/demo/src/features/mesh/jobs/write.rs
`FramePacket`, `SubmitPacket`: https://github.com/DavidVonDerau/rafx/blob/dvd/jobs/demo/src/features/mesh/internal/frame_packet.rs

# Replace nodes with render_features and new jobs framework.
401a3e5 

This is an alternative approach to handling the `FramePacket`, `ViewPacket`, and `SubmitNode` construction. I have tried to closely match the design of the Destiny slides[1]. Each feature now defines custom data for allocation in the frame packet, view packet, and submit packets explicitly. In order to avoid a proliferation of generics, there are type-erased traits (indicated by a `RenderFeature` prefix) that hide the use of the generic types behind a `dyn` trait.

This work had two primary goals:
1. Features should not need to know about threads or tasks in order to be parallelized.
2. Writing a new feature should be as ergonomic as possible while maintaining the 1st requirement.

`Extract`, `Prepare`, and `Write` jobs now support the same entry points as those defined by Destiny, with the exception of the `per game object` entry points mentioned as perf & memory optimization. Each feature defines their data format in a `FramePacket` and `SubmitPacket`. 

The `FramePacket` is the data extracted from the world. The `SubmitPacket` is the data prepared for the `Write` job. Both the `FramePacket` and `SubmitPacket` are allocated up-front to minimize allocations and both are accessible to the `Write` job. 

Each `FramePacket` contains a `ViewPacket` for each `RenderView` and each `SubmitPacket` contains a `ViewSubmitPacket` for each `RenderView`. The `FramePacket` has a `RenderObjectInstance` for each entity and `RenderObject` in the current frame and the `ViewPacket` has a `RenderObjectInstancePerView` for each `RenderObjectInstance` in that particular `RenderView`'s visibility. The same layout follow the `SubmitPacket` and `ViewSubmitPacket`, but the `ViewSubmitPacket` also a pre-allocated list of `SubmitNodeBlock`s for each `RenderPhase` supported by that `RenderFeature`. 

All of the submit nodes from all features are unioned into `ViewPhaseSubmitNodeBlock`s according to a shared `RenderView` and `RenderPhase` and then sorted by the `RenderPhase`'s defined `SubmitNodeSortFunction`. 

image

A feature's `RenderObject`s may be stored in a `RenderObjectRegistry` -- this replaces the `NodeSet` previously used by the demo's `TileLayer`, `Sprite`, and `Mesh` render features.

The `Extract` entry points are defined by `ExtractJobEntryPoints` and the `Prepare` entry points are defined by `PrepareJobEntryPoints`. Each entry point receives a `Context` argument (e.g. `ExtractPerFrameContext`) to minimize the loss of performance for unneeded data by that render feature. References to the `World` or other game resources can be cached while creating the `Extract`, `Prepare`, or `Write` jobs. Each job type is tied to an explicit lifetime (e.g. `'extract`) representing that stage of the renderer pipeline.

[1] https://advances.realtimerendering.com/destiny/gdc_2015/Tatarchuk_GDC_2015__Destiny_Renderer_web.pdf

##  Add new collections to rafx-base.
ca335df

- An `OwnedPool<T>` is a lightweight pool that moves a `Pooled<T>` to the user. `Pooled<T>` implements `Deref` and `DerefMut` for `T`. When the `Pooled<T>` is dropped, the `T` is moved back to the `OwnedPool<T>` via a channel.
- Added a new key called `RawDropSlabKey` for the drop slab. The `RawDropSlabKey` is a type-erased variant of `DropSlabKey` so that it can be stored as a handle in a type or collection without requiring a generic `T`.

There are 3 variants of a generic thread-safe storage that don't require default initialization.

-  `AtomicOnceCell` is a thread-safe variant of `OnceCell`.
-  `AtomicOnceCellArray` is an indexed variant representing a contiguous fixed-size array of the cells using an atomic bitvector to track accesses.
-  `AtomicOnceCellStack` is built on top of `AtomicOnceCellArray` and provides `push` / `reserve` / `set` semantics using an atomic to track the top of the stack.

All of the `Atomic` storages will `panic` if used incorrectly, e.g. by trying to set the same cell simultaneously from multiple threads, or by attempting to read a cell that has not yet been initialized via a call to `set` or `push`. `AtomicOnceCell` and `AtomicOnceCellArray` contain tests but `AtomicOnceCellStack` does not because the safety invariants of the `AtomicOnceCellStack` rely on the underlying `AtomicOnceCellArray`.

## Update Renderer to use new render_features code.
cfeae1a

- The `RenderFeaturePlugin` has been expanded with additional functions for calculating and allocating frame or submit packets, creating each job type, and determining if a given `RenderView` and `RenderViewVisibilityQuery` is relevant to that feature.
- Added `RendererThreadPool` to allow the application to control the parallelization of different stages of the renderer pipeline. `RendererThreadPoolNone` is provided as a single-threaded default if the application does not provide their own implementation.
- Seperated `RendererAssetPlugin` from what was previously just a `RendererPlugin` containing both feature code (now moved to `RenderFeaturePlugin`) and asset code.
- Added many helper functions to `Renderer` and `RenderFrameJob` to make the implementation of the `RendererThreadPool` easier.

## Replace nodes_api_design with renderer_triangle example.
9d11f1f

This example builds on the `asset_triangle` example to show how to use `rafx-renderer` and a `RenderFeaturePlugin` to drive rendering.

##  Update the demo to use the new render_features code.
320909a

- Every feature has been rewritten to use the new `extract` and `prepare` entry points and to define their custom `FramePacket` or `SubmitPacket` data.
- The demo includes an example implementation of `RendererThreadPool` using `bevy-tasks`.


rafx-base/src/atomic_once_cell.rs
```diff
@@ -0,0 +1,434 @@
...
+            data: MaybeUninit::uninit(),
+            is_initialized: AtomicUsize::new(0),
+        }
+    }
+
+    #[inline(always)]
+    fn start_set(&self) {
+        match self
+            .is_initialized
+            .fetch_update(Ordering::AcqRel, Ordering::Relaxed, |atomic_val| {
```
aclysma, line 27: Comments to justify the ordering chosen here would be good

rafx-base/src/atomic_once_cell.rs
```diff
@@ -0,0 +1,434 @@
...
+            }
+            _ => unreachable!(),
+        };
+    }
+
+    #[inline(always)]
+    fn end_set(&self) {
+        match self
+            .is_initialized
+            .fetch_update(Ordering::Release, Ordering::Relaxed, |atomic_val| {
```
aclysma, line 44: Comments to justify the ordering chosen here would be good

rafx-base/src/atomic_once_cell.rs
```diff
@@ -0,0 +1,434 @@
...
+pub struct AtomicOnceCell<T> {
+    data: MaybeUninit<UnsafeCell<T>>,
+    is_initialized: AtomicUsize,
+}
+
+impl<T> AtomicOnceCell<T> {
+    pub fn new() -> Self {
+        Self {
+            data: MaybeUninit::uninit(),
+            is_initialized: AtomicUsize::new(0),
```
aclysma, line 19: Should this be a full usize or smaller?

rafx-base/src/atomic_once_cell.rs
```diff
@@ -0,0 +1,434 @@
+use core::ptr;
+use std::cell::UnsafeCell;
+use std::mem::MaybeUninit;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+const SET_ACQUIRE_FLAG: usize = 1 << 1;
+const SET_RELEASE_FLAG: usize = 1 << 0;
+const IS_INIT_BITMASK: usize = SET_ACQUIRE_FLAG | SET_RELEASE_FLAG;
+
+pub struct AtomicOnceCell<T> {
```
aclysma, line 10: I imagine you already have this in your mental TODOs but we should summarize the purpose of these containers and what problem they are meant to solve in the higher level code

rafx-base/src/owned_pool.rs
```diff
@@ -0,0 +1,91 @@
+use crossbeam_channel::{bounded, Receiver, Sender};
+use std::ops::{Deref, DerefMut};
+
+pub struct OwnedPool<T> {
```
aclysma, line 4: I feel like we already had something like this but it was probably GPU-specific and waiting N frames until clearing a resource and allowing them to be reused. Maybe it was removed.

rafx-base/src/slab/drop_slab.rs
```diff
@@ -126,6 +58,14 @@ impl<T> DropSlab<T> {
         DropSlabKey::new(slab_key, self.drop_tx.clone())
     }
 
+    pub fn allocate_raw(
```
aclysma, line 61: What is this used for? Do we need one slab implementation to support both? (Wondering if it's cleaner to have two separate slab types.)

rafx-framework/Cargo.toml
```diff
@@ -31,6 +31,9 @@ serde_bytes = "0.11"
 glam = { version = "0.13.1", features = [ "serde" ] }
 parking_lot = "0.11.1"
 slotmap = "1.0.2"
+downcast-rs = "1.2.0"
+dashmap = "4.0.2"
```
aclysma, line 35: dashmap makes a bucket per CPU core and some CPUs have lots of cores, so we should be highly selective about how/where we use it. How are we using it in this change?

rafx-framework/Cargo.toml
```diff
@@ -31,6 +31,9 @@ serde_bytes = "0.11"
...
+rustc-hash = "1.1.0"
```
aclysma, line 36: I think we should try to use FNV consistently everywhere (or something else consistently everywhere) unless there is extremely good reason to add more of them.

rafx-framework/src/render_features/jobs/extract/extract_job.rs
```diff
@@ -0,0 +1,210 @@
...
+
+struct ExtractScopeNames {
+    begin_per_frame_extract: String,
+    extract_render_object_instance: String,
+    extract_render_object_instance_per_view: String,
+    end_per_view_extract: String,
+    end_per_frame_extract: String,
+}
+
+impl ExtractScopeNames {
```
aclysma, line 190: Are we having to do the string formats every time we create the job? lazy_static might let us just do it once and reuse.

rafx-framework/src/render_features/jobs/extract/render_feature_extract_job.rs
```diff
@@ -0,0 +1,43 @@
...
+        view_packet: &dyn RenderFeatureViewPacket,
+        range: Range<usize>,
+    );
+
+    fn end_per_view_extract(
+        &self,
+        view_packet: &dyn RenderFeatureViewPacket,
+    );
+
+    fn end_per_frame_extract(&self);
```
aclysma, line 32: is there documentation for these elsewhere? if not, I'm sure it's something you already planned to add :)

rafx-framework/src/render_features/jobs/extract/render_feature_view_packet.rs
```diff
@@ -0,0 +1,62 @@
...
+        render_object_instance: RenderObjectInstance,
+    ) -> RenderObjectInstancePerViewId;
+
+    fn push_volume(
+        &mut self,
+        entity_id: EntityId,
+    );
+}
+
+pub trait RenderFeatureViewPacketIntoConcrete {
```
aclysma, line 32: What is this for? (comments would be good, again I'm sure you had it on your mental TODO list :) )

rafx-framework/src/render_features/jobs/prepare/prepare_job.rs
```diff
@@ -0,0 +1,257 @@
...
+    begin_per_frame_prepare: String,
+    prepare_render_object_instance: String,
+    prepare_render_object_instance_per_view: String,
+    end_per_view_prepare: String,
+    end_per_frame_prepare: String,
+}
+
+impl PrepareScopeNames {
+    pub fn new(feature_name: &str) -> Self {
+        Self {
```
aclysma, line 239: maybe we can lazy_static this

aclysma: or create these strings when we register the feature?

rafx-framework/src/render_features/jobs/prepare/render_feature_prepare_job.rs
```diff
@@ -0,0 +1,49 @@
+use crate::render_features::render_features_prelude::*;
+use std::ops::Range;
+
+/// TODO(dvd): Document.
```
aclysma, line 4: 👍

rafx-framework/src/visibility/entity_id.rs
```diff
@@ -1,3 +1,8 @@
+use std::hash::Hash;
+
+/// An opaque 64-bit handle used as a unique identifier for objects in the game world
+/// that are visible or otherwise relevant to the renderer's pipeline. Each `VisibilityObject`
+/// is associated with a specific `EntityId`.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct EntityId(u64);
```
aclysma, line 7: I'd consider making this UserId or VisibilityId (can be a separate change!)

rafx-framework/src/visibility/visibility_object_allocator.rs
```diff
@@ -26,6 +26,19 @@ impl<'a> Deref for VisibilityObjectRef<'a> {
     }
 }
 
+pub struct VisibilityObjectLookup<'a>(
+    RwLockReadGuard<'a, SlotMap<VisibilityObjectId, VisibilityObject>>,
```
aclysma, line 30: How often would we be hitting this lock?

rafx-renderer/src/render_frame_job.rs
```diff
@@ -1,27 +1,25 @@
...
-use rafx_framework::nodes::{
-    FramePacket, PrepareJobSet, RenderJobPrepareContext, RenderRegistry, RenderView,
-};
+use rafx_framework::render_features::render_features_prelude::*;
 use rafx_framework::{DynCommandBuffer, RenderResources, ResourceContext};
 use std::sync::Arc;
 
 pub struct RenderFrameJobResult;
 
 pub struct RenderFrameJob {
```
aclysma, line 13: docs for what this does

rafx-renderer/src/renderer_thread_pool_none.rs
```diff
@@ -0,0 +1,149 @@
+use crate::{RenderFeaturePlugin, RenderFrameJob, Renderer, RendererThreadPool};
+use rafx_framework::render_features::render_features_prelude::*;
+use std::sync::Arc;
+
+#[derive(Clone)]
```
aclysma, line 5: Comments explaining that this is a "default" pool that doesn't do multithreading, and that we expect end-users to provide their own pool, which allows them to dispatch work into whatever job system/thread pool that they have.

rafx/examples/renderer_triangle/features/jobs/extract.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+impl<'extract> ExtractJobEntryPoints<'extract> for DemoExtractJob<'extract> {
+    fn begin_per_frame_extract(
+        &self,
+        _render_objects: &RenderObjectsMap<DemoRenderObjectStaticData>,
+        context: &ExtractPerFrameContext<'extract, '_, Self>,
+    ) {
+        context
```
aclysma, line 44: I think having an example that sets up the jobs is a great idea. Maybe an extremely basic particle system where we extract some positions, produce some sprites around each position, and then draw them. This way we use a little more of the job system and show how per-entity stuff works. This can be a separate thing we do in the future. Happy for this to go in as it is now.

rafx/examples/renderer_triangle/renderer_triangle.rs
```diff
@@ -0,0 +1,375 @@
...
+        let (width, height) = sdl2_window.vulkan_drawable_size();
+        let swapchain_helper = SwapchainHandler::create_swapchain(
+            &mut renderer_builder_result.asset_manager,
+            &mut renderer_builder_result.renderer,
+            sdl2_window,
+            width,
+            height,
+        )?;
+
+        resources.insert(api.device_context());
```
aclysma, line 145: We should add lots of comments to this file to explain what all these steps are and why they're necessary/helpful

aclysma:
Great work! I think we should pick just a few of the "most important suggestions" now (like the addition of the rusts-hash crate as a dependency) and follow up on most of the feedback after merging. Almost all of it is very minor/optional suggestion or questions, and I want to make sure we merge this before any other changes so that we avoid merge conflicts on such a huge change.

rafx-base/src/owned_pool.rs
DavidVonDerau: Yes, the `PooledResourceAllocator`. We still have it & you identified the difference -- it waits `N` frames until the resource is cleared & allowed to be reused. Also, I'm not sure it is actually `pub` outside of the `rafx` code that uses it. The function signature for creating a new resource is also different -- it assumes the resource is created by a `RafxDeviceContext`.

rafx-framework/Cargo.toml
DavidVonDerau: In my testing, `rustc-hash` has been significantly faster than `FNV`, e.g.
```
// 5.3909 milliseconds to query the world for 150000 entities using 1 parallel `query_visibility`. (std::hashmap)
// ~4.5 milliseconds to query the world for 150000 entities using 1 parallel `query_visibility`. (fnvhashmap)
// 2.3005 milliseconds to query the world for 150000 entities using 1 parallel `query_visibility`. (identity hash)
// 2 milliseconds to query the world for 150000 entities using 1 parallel `query_visibility`. (fxhashmap)
```
My vote is therefore with `rustc-hash`.

rafx-framework/src/visibility/visibility_object_allocator.rs
DavidVonDerau: Inside of `query_visibility` in the `ViewVisibilityJob`, every single `VisibilityResult` must be mapped back to a `VisibilityObject` to figure out what `RenderObject`s are associated with it. So this guard is created at the top of `query_visibility` and however many thousands of look-ups occur all resolve against the already acquired lock instead of needing a new lock each time.

rafx-renderer/src/renderer_thread_pool_none.rs
DavidVonDerau: This pool probably shouldn't be `pub` since it's created by the `Renderer` if the user passed `None` for the `RendererThreadPool` parameter. If it's not `pub`, do you still want the comment?

The main reason why I'm asking is because `RendererThreadPool` (the trait) _does_ have an extensive comment on it.

rafx-framework/src/render_features/jobs/extract/render_feature_extract_job.rs
DavidVonDerau: Yes, specifically in `ExtractJobEntryPoints` (and `RenderFeatureExtractJob` does have a comment that points users to look at the entry points instead, ditto for `ExtractJob`)

rafx-framework/src/visibility/entity_id.rs
DavidVonDerau: Will defer for now.

demo/src/features/tile_layer/jobs/prepare.rs
```diff
@@ -1,92 +1,100 @@
...
-use rafx::framework::{MaterialPassResource, ResourceArc};
-
-/// Per-pass "global" data
-pub type TileLayerUniformBufferObject = shaders::tile_layer_vert::ArgsUniform;
+use rafx::framework::DescriptorSetAllocatorRef;
 
 pub struct TileLayerPrepareJob {
-    visible_render_nodes: Vec<TileLayerRenderNode>,
-    tile_layer_material: ResourceArc<MaterialPassResource>,
+    descriptor_set_allocator: Mutex<DescriptorSetAllocatorRef>,
```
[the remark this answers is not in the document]
DavidVonDerau: Removed.

demo/src/features/text/jobs/extract.rs
```diff
@@ -1,58 +1,69 @@
...
 use fnv::FnvHashMap;
-use rafx::assets::AssetManagerRenderResource;
+use rafx::assets::{AssetManagerRenderResource, MaterialAsset};
+use rafx::base::resource_map::ReadBorrow;
+use rafx::base::resource_ref_map::ResourceRefBorrowMut;
+use rafx::distill::loader::handle::Handle;
 
-pub struct TextExtractJob {}
+pub struct TextExtractJob<'extract> {
+    text_resource: Mutex<ResourceRefBorrowMut<'extract, TextResource>>,
```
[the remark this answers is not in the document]
DavidVonDerau: Replaced with `TrustCell`.

demo/src/features/mesh/shadow_map_resource.rs
```diff
@@ -6,23 +6,22 @@ use crate::phases::ShadowMapRenderPhase;
...
+    RenderPhaseMaskBuilder, RenderView, RenderViewDepthRange, RenderViewSet,
+};
+use rafx::visibility::{EntityId, ViewFrustumArc};
 
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub enum LightId {
-    PointLight(legion::Entity), // u32 is a face index
-    SpotLight(legion::Entity),
-    DirectionalLight(legion::Entity),
+    PointLight(EntityId), // u32 is a face index
```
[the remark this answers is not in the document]
DavidVonDerau: Resolving as dup of https://github.com/aclysma/rafx/pull/135#discussion_r632188070/

demo/src/features/mesh/jobs/prepare.rs
```diff
@@ -74,186 +41,98 @@ struct PreparedSpotLight<'a> {
...
 
-pub struct MeshPrepareJob {
-    depth_material: ResourceArc<MaterialPassResource>,
-    extracted_frame_node_mesh_data: Vec<Option<ExtractedFrameNodeMeshData>>,
-    directional_lights: Vec<ExtractedDirectionalLight>,
-    point_lights: Vec<ExtractedPointLight>,
-    spot_lights: Vec<ExtractedSpotLight>,
+pub struct MeshPrepareJob<'prepare> {
+    resource_context: ResourceContext,
+    descriptor_set_allocator: Mutex<DescriptorSetAllocatorRef>,
```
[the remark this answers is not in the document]
DavidVonDerau: Removed.

demo/src/features/debug3d/jobs/extract.rs
```diff
@@ -1,46 +1,59 @@
...
-        Self {}
+impl<'extract> Debug3DExtractJob<'extract> {
+    pub fn new(
+        extract_context: &RenderJobExtractContext<'extract>,
+        frame_packet: Box<Debug3DFramePacket>,
+        debug3d_material: Handle<MaterialAsset>,
+    ) -> Arc<dyn RenderFeatureExtractJob<'extract> + 'extract> {
+        Arc::new(ExtractJob::new(
+            Self {
+                debug3d_resource: Mutex::new(
```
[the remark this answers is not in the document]
DavidVonDerau: Borrow checker. Replaced with `TrustCell`.

demo/src/features/debug3d/jobs/prepare.rs
```diff
@@ -1,99 +1,79 @@
...
-use rafx::framework::{MaterialPassResource, ResourceArc};
-
-pub type Debug3dUniformBufferObject = shaders::debug_vert::PerFrameUboUniform;
+use rafx::api::{RafxBufferDef, RafxDeviceContext, RafxMemoryUsage, RafxResourceType};
+use rafx::framework::{DescriptorSetAllocatorRef, DynResourceAllocatorSet};
 
 pub struct Debug3DPrepareJob {
-    debug3d_material_pass: ResourceArc<MaterialPassResource>,
-    line_lists: Vec<LineList3D>,
+    descriptor_set_allocator: Mutex<DescriptorSetAllocatorRef>,
```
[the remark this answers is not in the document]
DavidVonDerau: Removed.

demo/src/features/debug3d/plugin.rs
```diff
@@ -1,26 +1,55 @@
...
+    pub fn legion_init(
+        &self,
+        resources: &mut legion::Resources,
+    ) {
+        resources.insert(Debug3DResource::new());
     }
 
     pub fn legion_destroy(resources: &mut legion::Resources) {
-        resources.remove::<DebugDraw3DResource>();
+        resources.remove::<Debug3DResource>();
```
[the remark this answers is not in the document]
DavidVonDerau: The name was changed to match the feature, yeah. I would prefer to make a change like that in a separate PR, e.g. just `rename Debug3D to DebugDraw3D`.

demo/src/features/imgui/internal/imgui_draw_data.rs
```diff
@@ -37,6 +39,7 @@ impl ImGuiDrawList {
     }
 }
 
+#[derive(Clone)]
```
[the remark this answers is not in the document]
DavidVonDerau: Nope, not sure why this was added. Removed.

demo/src/features/imgui/jobs/extract.rs
```diff
@@ -1,79 +1,89 @@
...
+use super::*;
+use rafx::assets::{AssetManagerRenderResource, MaterialAsset};
+use rafx::base::resource_map::ReadBorrow;
+use rafx::base::resource_ref_map::ResourceRefBorrowMut;
+use rafx::distill::loader::handle::Handle;
 use rafx::graph::SwapchainSurfaceInfo;
 
-pub struct ImGuiExtractJob {}
+pub struct ImGuiExtractJob<'extract> {
+    sdl2_imgui_manager: Mutex<ResourceRefBorrowMut<'extract, Sdl2ImguiManager>>,
```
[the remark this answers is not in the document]
DavidVonDerau: Now it's a `TrustCell`.

demo/src/features/imgui/jobs/prepare.rs
```diff
@@ -1,88 +1,91 @@
...
+use rafx::api::{RafxBufferDef, RafxDeviceContext, RafxMemoryUsage, RafxResourceType};
+use rafx::framework::{
+    DescriptorSetAllocatorRef, DynResourceAllocatorSet, ImageViewResource, ResourceArc,
+};
 
 pub struct ImGuiPrepareJob {
-    imgui_draw_data: Option<ImGuiDrawData>,
-    imgui_material_pass: ResourceArc<MaterialPassResource>,
-    view_ubo: ImGuiUniformBufferObject,
+    descriptor_set_allocator: Mutex<DescriptorSetAllocatorRef>,
```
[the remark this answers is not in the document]
DavidVonDerau: Removed.

demo/src/features/imgui/jobs/write.rs
```diff
@@ -21,110 +25,117 @@ lazy_static::lazy_static! {
...
-        write_context: &mut RenderJobWriteContext,
-        _view: &RenderView,
+        write_context: &mut RenderJobCommandBufferContext,
+        _view_frame_index: ViewFrameIndex,
         _render_phase_index: RenderPhaseIndex,
-        index: SubmitNodeId,
+        submit_node_id: SubmitNodeId,
     ) -> RafxResult<()> {
-        profiling::scope!(super::RENDER_ELEMENT_SCOPE_NAME);
+        profiling::scope!(super::RENDER_SUBMIT_NODE_SCOPE_NAME);
```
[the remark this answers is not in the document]
DavidVonDerau: Technically I could put it in `write_view_phase` above the call to this, but I'd prefer to _not_ have N scopes for N draw calls. I'd rather just profile the time in each feature. That would require a manual `start` and `end` scope function on `profiling` though.

demo/src/render_graph_generator/ui_pass.rs
```diff
@@ -38,8 +38,11 @@ pub(super) fn ui_pass(
     // buffer. Just add the draw calls.
     let main_view = context.main_view.clone();
     context.graph.set_renderpass_callback(node, move |args| {
+        profiling::scope!("UI Pass");
```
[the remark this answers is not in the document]
DavidVonDerau: Yeah, I think that would be cool. These scopes were pretty helpful to add.

rafx-base/src/atomic_once_cell.rs
DavidVonDerau: Changed to `AtomicU8`.

rafx-framework/src/render_features/jobs/extract/render_feature_view_packet.rs
DavidVonDerau: It's to make this type of syntactic sugar work: 
image

If I remember correctly, I had trouble getting the `into` trait to work with the generics / downcasting, so this was my workaround.

rafx-framework/Cargo.toml
DavidVonDerau: It's used in two places during extract.
1. There's some per-feature work that would generate a bunch of allocations if not reused frame-to-frame, so those are put into a `DashMap` indexed by the `RenderFeatureId`. This could be replaced, I think, by a `Vec` of `TrustCell`.
2. It's used to hold the temporary `Vec`s containing the entity / render object pairs after they've been categorized by feature & sorted. This is indexed by a persistent ID tied to the `RenderView`'s `ViewFrustum`. I'm not sure if there is a good way to replace this usage.

demo/src/demo_renderer_thread_pool.rs
```diff
@@ -0,0 +1,399 @@
...
+#[derive(Default)]
+struct ParallelChunkSizes {
+    extract_chunk_size: Option<usize>,
+    extract_per_view_chunk_size: Option<usize>,
+    prepare_chunk_size: Option<usize>,
+    prepare_per_view_chunk_size: Option<usize>,
+}
+
+impl ParallelChunkSizes {
+    pub fn extract_chunk_size(
```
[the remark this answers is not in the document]
DavidVonDerau: It's just set of "fluent" functions for configuring the chunk sizes. So it's builder-ish, but unlike a builder there's no final `build` function since each function is mutating the thing you'd return anyways. It's also not passed anywhere else -- it's only used inside of the `DemoRendererThreadPool` to demonstrate an example of how a user can do arbitrary parallelization. I thought that was covered by the comment inside of `DemoRendererThreadPool::new()`.

I'm moving this to the bottom of the file so that a reader sees the other comment first.

demo/src/phases/depth_prepass_render_phase.rs
```diff
@@ -1,5 +1,5 @@
-use rafx::nodes::RenderPhase;
-use rafx::nodes::{RenderPhaseIndex, SubmitNode};
+use rafx::render_features::RenderPhase;
```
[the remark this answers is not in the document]
DavidVonDerau: I like `render_features` since `RenderFeature` is already being used as a prefix on a bunch of things.

rafx-base/src/atomic_once_cell.rs
DavidVonDerau: Is the documentation on `AtomicOnceCellArray` and `AtomicOnceCellStack` acceptable? If so, I can write a similar doc comment on `AtomicOnceCell`. If you want something more specific / different, can you give an example?

rafx-base/src/slab/drop_slab.rs
DavidVonDerau: A `RenderObjectHandle` is a reference-counted handle for a `RenderObject` associated with some `RenderFeature`. The `RenderObject`s are allocated out of some `DropSlab`. We'd like the `RenderObjectHandle` to contain the key so that 1.) it can be used for the look-up, and 2.) when there are no more references to the `RenderObjectHandle`, the `DropSlab` can free that slot. However, we also want to put `RenderObjectHandle`s into lists or hash maps or other containers. If the `DropSlabKey` includes the generic `T`, we can't do that because the different `T` arguments will bleed into the `RenderObjectHandle`. 

So, this allocates a key that is _not_ generic on `T`. Other than that it's the exact same as `DropSlabKey`. If you think back to the code today, it's basically the same as the `GenericNodeRenderHandle` if the `GenericNodeRenderHandle` is what truly represented "ownership" of the entry in the slab that allocated nodes.

I would maybe lean towards nixing the `T` from `DropSlabKey` and then getting rid of the key type that I added? If you're concerned about keys allocated from one `T` being used in a different `T`, we could include the `TypeId` and assert on that in the various `get` functions, or do some type of equality check using the `Sender` on the key (assuming we can compare it against our `Receiver`?)

rafx-framework/src/render_features/jobs/extract/extract_job.rs
DavidVonDerau: How does `lazy_static` work if the string formats depend on the feature name?

rafx-framework/src/render_features/jobs/prepare/prepare_job.rs
DavidVonDerau: Creating them when the feature is registered is an interesting idea. I can look at that.

rafx-renderer/src/render_frame_job.rs
DavidVonDerau: Added a `TODO(dvd): DOCUMENT` to track that.

rafx/examples/renderer_triangle/features/jobs/extract.rs
DavidVonDerau: Agree with making a more detailed example later. This example is also just _incredibly_ helpful when making changes to the jobs. Only having to update a single feature is a life saver on iteration times.

rafx/examples/renderer_triangle/renderer_triangle.rs
DavidVonDerau: Is that a TODO for now, or a TODO for later? :philosorator:

demo/src/components.rs
```diff
@@ -1,18 +1,17 @@
...
 use glam::f32::Vec3;
 use glam::Quat;
 use rafx::framework::visibility::VisibilityObjectArc;
+use rafx::render_features::RenderObjectHandle;
 use rafx::visibility::ViewFrustumArc;
 
 #[derive(Clone)]
 pub struct MeshComponent {
-    pub render_node: MeshRenderNodeHandle,
+    pub handle: RenderObjectHandle,
```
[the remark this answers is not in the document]
DavidVonDerau: Renamed.

demo/src/features/imgui/plugin.rs
```diff
@@ -71,12 +94,90 @@ impl RendererPlugin for ImGuiRendererPlugin {
...
+        )
+    }
+
+    fn new_submit_packet(
+        &self,
+        frame_packet: &Box<dyn RenderFeatureFramePacket>,
+    ) -> Box<dyn RenderFeatureSubmitPacket> {
+        let frame_packet: &ImGuiFramePacket = frame_packet.as_ref().as_concrete();
+
+        let mut view_submit_packets = Vec::with_capacity(frame_packet.view_packets().len());
```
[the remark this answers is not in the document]
DavidVonDerau: I am not sure that there is an easy way to do that. The submit node blocks are tied to 1.) the render phases the feature needs to create submit nodes in, and 2.) the number of submit nodes for each phase. First, features don't necessarily need the same number of submit node blocks (e.g. if they only create submit nodes for one render phase), but second, the number of submit nodes needed for each block is feature-dependent and can even vary between render phases (e.g. the mesh render feature doesn't need to create submit nodes in `Opaque` for any view used by a shadow caster).

There's probably a _trivial_ case that I can extract into a helper function -- something like, 1 render phase, max num submit nodes = max render object instances. I can see about that to reduce the duplication in any of the "simple" features.

demo/src/features/imgui/plugin.rs
```diff
@@ -30,7 +29,31 @@ impl ImGuiRendererPlugin {
...
+impl RenderFeaturePlugin for ImGuiRendererPlugin {
+    fn feature_debug_name(&self) -> &'static str {
+        super::render_feature_debug_name()
+    }
+
+    fn feature_index(&self) -> RenderFeatureIndex {
+        super::render_feature_index()
+    }
+
+    fn is_view_relevant(
```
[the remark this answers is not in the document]
DavidVonDerau: I'll look at the `Config` struct idea.

rafx-framework/Cargo.toml
aclysma: I'm happy to switch if the quality is still ok, I just think we should use FNV in this change and switch everything over as a separate change. I also think we should typedef something in rafx-base and change everything to use that, rather than changing all code to use a different crate directly.

rafx-renderer/src/renderer_thread_pool_none.rs
aclysma: I think we could do a really small comment like "This is a single-threaded default implementation that is used by default. End-users can override to integrate their own preferred method of scheduling work on threads."

rafx-framework/src/render_features/jobs/extract/render_feature_view_packet.rs
aclysma: Oh so this is essentially a downcast? Is it not possible to be a member of RenderFeatureViewPacket? If not possible, a short comment like "allows down-casting to a view packet of a specific feature" might be helpful.

demo/src/scenes/rafxmark_scene.rs
```diff
@@ -95,7 +99,7 @@ impl RafxmarkScene {
         let sprite_spawner_system = SystemBuilder::new("sprite_spawner")
             .read_resource::<TimeState>()
             .read_resource::<VisibilityRegion>()
-            .write_resource::<SpriteRenderNodeSet>()
+            .write_resource::<SpriteRenderObjectRegistry>()
```
aclysma, line 102: I generally use the word "registry" with things that get registered and stay forever, or at least a long time. Like the RenderRegistry where we permanently register features/phases. Think RenderObjectSet could be more clear that it's a set of objects that changes regularly. (And it's shorter)

aclysma: (I can do this as a follow-up change if you like)

rafx-base/src/atomic_once_cell.rs
aclysma: Something like what's on those two structs is fine. What I had in mind was something really short like "AtomicOnceCell is a value that can only be set once, using an atomic value to synchronize across threads rather than a mutex." Completely up to you, happy either way.

rafx-framework/src/render_features/jobs/extract/extract_job.rs
aclysma: yeah, I wasn't thinking through it fully. Maybe it would work if we have a lazy_static created per-feature with the declare_feature macro. I usually don't like doing things in macros, but I think if the alternative is string formatting per frame, I'd definitely consider it. The code might be a little like the RenderFeatureT::render_feature_index() function that returns a global atomic value. If it doesn't work, we can figure something else out.

aclysma: (Feel free to address later, or we can add an issue to try to improve later if you'd rather just leave it for now.)

rafx-framework/src/render_features/jobs/extract/render_feature_view_packet.rs
DavidVonDerau: It cannot be a member of `RenderFeatureViewPacket` (or any of the other type-erased traits) because it has a generic parameter (the return value). That prevents the trait from being made into an object and then that breaks everything else. The compiler advice is to put the function into a separate trait, which is I think how we got here. It's unfortunate though. It would be significantly cleaner the other way.

rafx-framework/src/render_features/jobs/prepare/prepare_job.rs
DavidVonDerau: Dupe of https://github.com/aclysma/rafx/pull/135#discussion_r631574648, will resolve there.

aclysma:
Went through anything that was changed, I trust your judgement on what should be addressed now, and what can be done later. If there's anything you think we should do later but you don't want to do right away (or don't want to do yourself), we can just add an issue for it. It's not a bad thing to have a few issues that are first-contributor-friendly, and some of the minor suggestions (like changing the hashing algorithm) would be good candidates for that.

rafx-framework/src/render_features/jobs/extract/extract_job.rs
DavidVonDerau: The best idea I've come up with so far is to replace the generated `feature_debug_name` in the macro with `feature_debug_constants` and point to a `const` `static` struct with the `static` strings for the feature name, scope names, etc.

And then if desired, `feature_debug_name()` can be reimplemented just as syntactic sugar around `feature_debug_constants().feature_name`.

demo/src/scenes/rafxmark_scene.rs
DavidVonDerau: Renamed.

rafx-framework/Cargo.toml
DavidVonDerau: Replaced with `fnv` for now.

rafx-renderer/src/renderer_thread_pool_none.rs
DavidVonDerau: Done.

demo/src/render_graph_generator/ui_pass.rs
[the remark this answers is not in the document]
DavidVonDerau: Opened issue https://github.com/aclysma/rafx/issues/136 to track

rafx/examples/renderer_triangle/features/jobs/extract.rs
DavidVonDerau: Tracking in https://github.com/aclysma/rafx/issues/137

rafx/examples/renderer_triangle/renderer_triangle.rs
DavidVonDerau: `TODO` for later -- tracking with https://github.com/aclysma/rafx/issues/137

rafx-framework/Cargo.toml
DavidVonDerau: https://github.com/aclysma/rafx/issues/138

rafx-base/src/atomic_once_cell.rs
DavidVonDerau: Added a comment & changed `AcqRel` to just `Acquire`
>  Use `Acquire` to start a protected section.

rafx-base/src/atomic_once_cell.rs
DavidVonDerau: Added a comment:
>  Use `Release` to end the protected section.

demo/src/features/imgui/plugin.rs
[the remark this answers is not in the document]
DavidVonDerau: Added a helper function.

rafx-base/src/slab/drop_slab.rs
DavidVonDerau: Removed `T` from `DropSlabKey`. When querying with the key, use `same_channel` to check that the key is from the right `DropSlab`.

rafx-framework/src/render_features/jobs/extract/render_feature_view_packet.rs
DavidVonDerau: Added comments.

rafx-framework/Cargo.toml
DavidVonDerau: Removed all uses in the first category.

demo/src/features/imgui/plugin.rs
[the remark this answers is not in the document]
aclysma: If it doesn't make things considerably simpler, I'm in favor of leaving it as-is

rafx-base/src/slab/drop_slab.rs
aclysma: I see, if we need to have typed/untyped handles, I would have the API always use typed handles, but allow converting typed -> untyped and untyped -> typed. If it's painful to do the explicit conversion back to typed (because it requires generics where we don't want them) then just asserting on typeid is fine.

rafx-framework/Cargo.toml
aclysma: I'd lean toward using a plain mutex until a profiler shows that as being a bottleneck.

rafx-framework/src/render_features/jobs/extract/extract_job.rs
aclysma: That sounds good. We might could also have the macro produce functions returning 'static &str for each of these. I do like grouping these under something like feature_debug_constants() to clearly separate what shouldn't be relied upon for correctness.

demo/src/features/imgui/plugin.rs
```diff
@@ -30,7 +29,26 @@ impl ImGuiRendererPlugin {
     }
 }
 
-impl RendererPlugin for ImGuiRendererPlugin {
```
aclysma, line 33: Should this be ImGuiRenderFeaturePlugin now?

aclysma: (lets do any renames after we merge this)

rafx-framework/src/render_features/jobs/prepare/view_phase_submit_node_block.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// Example: https://web.archive.org/web/20210110113523/https://realtimecollisiondetection.net/blog/?p=86
+pub type SubmitNodeSortKey = u32;
+
+/// The sort function used by a particular `RenderPhase`. This is usually one of the following:
+/// 1. front-to-back
+/// 2. back-to-front
+/// 3. by feature index
+/// 4. unsorted
+pub type SubmitNodeSortFunction = fn(&mut Vec<RenderFeatureSubmitNode>);
```
aclysma, line 19: Maybe one day this would be an enum with a few pre-implemented options + a callback for doing something custom

docs/renderer/renderer_architecture.md
```diff
@@ -0,0 +1,345 @@
+# Renderer Architecture
+
+`rafx-renderer` and the `rafx-framework` `render_features` were inspired by the 2015 GDC talk "[Destiny's Multithreaded Rendering Architecture](http://advances.realtimerendering.com/destiny/gdc_2015/Tatarchuk_GDC_2015__Destiny_Renderer_web.pdf)".
```
aclysma, line 3: I'll probably read through this in more detail later

aclysma (approved):

PR #139, opened 2021-05-15
GLES2 improvements:

aclysma:
- FBOs render with inverted Y so that they can be sampled in the same way as other backends
 - Fix textures bound in descriptor sets using wrong texture unit index
 - Support debug output in GL
 - Demo now uses depth images for the default "invalid" shadow maps


rafx-assets/src/assets/graphics_pipeline/assets.rs
```diff
@@ -247,7 +247,7 @@ impl MaterialPassData {
                 return Err(error)?;
             }
 
-            log::trace!("  Reflection data:\n{:#?}", reflection_data);
+            //log::trace!("  Reflection data:\n{:#?}", reflection_data);
```
DavidVonDerau, line 250: Commented out?

rafx-framework/src/resources/cooked_shader.rs
```diff
@@ -100,7 +100,7 @@ impl ReflectedShader {
         // We iterate through the entry points we will hit for each stage. Each stage may define
         // slightly different reflection data/bindings in use.
         for reflection_data in entry_points {
-            log::trace!("  Reflection data:\n{:#?}", reflection_data);
+            //log::trace!("  Reflection data:\n{:#?}", reflection_data);
```
DavidVonDerau, line 103: Also commented.

demo-web/shaders/shader.vert.gles
```diff
@@ -15,5 +15,7 @@ void main()
 {
     interface_var_0 = in_uv;
     gl_Position = vec4(pos, 0.0, 1.0);
```
DavidVonDerau, line 17: What's going on here? Is this how people would have to write GL shaders? Or is this the generated shader code?

DavidVonDerau (approved):
Just a few questions, otherwise looks good.

demo-web/shaders/shader.vert.gles
aclysma: the shader processor inserts this fix automatically

rafx-assets/src/assets/graphics_pipeline/assets.rs
aclysma: yeah these are super spammy, even for trace

PR #142, opened 2021-05-16
GL ES 3.0 shader support

aclysma:
There is a trivial CI change that needs rebased


DavidVonDerau:
Some of the checks failed?

DavidVonDerau (approved):

PR #143, opened 2021-05-16
Support additional render options

DavidVonDerau:
Support additional render options.

- Wireframes are drawn in the `WireframeRenderPhase` after the `OpaqueRenderPhase` and `TransparentRenderPhase`.
- The `Debug3D` feature now draws into the `WireframeRenderPhase`.
- The `MeshFeature` can create submit nodes in the `WireframeRenderPhase` if the `RenderView` is registered for the `MeshWireframeRenderFeatureFlag`.
- Added `show_wireframes` bool to ImGui settings to register the main view with the `MeshWireframeRenderFeatureFlag`.
- Added `show_surfaces` bool to ImGui settings to skip the `write_view_phase` for `OpaqueRenderPhase` and `TransparentRenderPhase`.
- Added `enable_lighting` bool to ImGui settings to enable or disable PBR lighting calculations.
- Set shadow map index to -1 when `show_shadows` is false. This skips the `calculate_percent_lit` code entirely.
- Added `show_textures` bool to ImGui settings to enable or disable the PBR textures.
- Reduced PCF samples to 9.

image
image
image
image
image


demo/assets/materials/debug.material
```diff
@@ -1,11 +1,11 @@
 (
     passes: [
         (
-            phase: Some("Opaque"),
+            phase: Some("Wireframe"),
             fixed_function_state: (
                 alpha_blending: Enabled,
                 depth_testing: EnabledReverseZ,
-                polygon_mode: Some(Line)
+                fill_mode: Some(Wireframe)
```
aclysma, line 8: I wonder why this never threw an error before

demo/src/lib.rs
```diff
@@ -139,6 +139,8 @@ pub struct RenderOptions {
     pub enable_msaa: bool,
     pub enable_hdr: bool,
     pub enable_bloom: bool,
+    pub show_models: bool,
```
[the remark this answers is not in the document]
aclysma: I think we should have the word "mesh" in all these things that are changing how the mesh feature works.

demo/src/features/mesh/mod.rs
```diff
@@ -6,8 +6,42 @@ use internal::*;
...
+    /// The `MeshWireframeRenderFeature` is a NOP if the view is not registered for the `MeshRenderFeature`.
+    use rafx::render_feature_mod_prelude::*;
+    rafx::declare_render_feature!(MeshWireframeRenderFeature, MESH_WIREFRAME_FEATURE_INDEX);
+}
+
+mod mesh_basic {
+    /// The `Mesh` feature generates untextured submit nodes if the `MeshBasicRenderFeature` is registered
+    /// on the view. The `MeshBasicRenderFeature` is a NOP if the view is not registered for the `MeshRenderFeature`.
+    use rafx::render_feature_mod_prelude::*;
+    rafx::declare_render_feature!(MeshBasicRenderFeature, MESH_BASIC_FEATURE_INDEX);
```
aclysma, line 21: Maybe it's more clear to call it unlit? (I'm being a little picky with names because we're getting enough things that it may get harder to keep track of them.)

demo/src/scenes/shadows_scene.rs
```diff
@@ -343,6 +349,24 @@ fn update_main_view_3d(
...
+        feature_mask_builder =
+            feature_mask_builder.add_render_feature::<MeshWireframeRenderFeature>();
+    }
+
+    if !render_options.enable_lighting {
+        feature_mask_builder = feature_mask_builder.add_render_feature::<MeshUnlitRenderFeature>();
+    }
+
+    if !render_options.enable_textures {
+        feature_mask_builder = feature_mask_builder.add_render_feature::<MeshBasicRenderFeature>();
```
aclysma, line 362: MeshUntexturedRenderFeature?

demo/shaders/glsl/pbr_frag.glsl
```diff
@@ -0,0 +1,968 @@
+layout (location = 0) in vec3 in_position_vs;
```
aclysma, line 1: I think it would be clearer if we keep all the shaders related to rendering the mesh feature prefixed with "mesh_"

I think this works well with most of the filenames (mesh_basic, mesh_wireframe). mesh_pbr would make more sense to me.

I also think we should avoid "basic" because it's very generic. Are there any changes other than not textured? It's still pulling in pbr. So maybe "mesh_pbr_untextured" and "mesh_pbr" would be accurate but succinct?

aclysma (commented):
Added comments

demo/assets/materials/debug.material
DavidVonDerau: I thought it's because the struct is defined as `Option` so it was just defaulting to `None`. Unless you thought that unexpected fields would cause the parser to fail?

aclysma: yeah, ideally they would. they're almost always a mistake

rafx-renderer/src/render_frame_job.rs
```diff
@@ -195,9 +195,12 @@ impl RenderFrameJob {
         prepare_context: &RenderJobPrepareContext<'prepare>,
         frame_packet: Box<dyn RenderFeatureFramePacket>,
     ) -> Arc<dyn RenderFeaturePrepareJob<'prepare> + 'prepare> {
-        let feature = features.get(frame_packet.feature_index() as usize).unwrap();
-        profiling::scope!(feature.feature_debug_constants().feature_name);
+        let feature = features
```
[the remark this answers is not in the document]
DavidVonDerau: Fixed.

demo/src/lib.rs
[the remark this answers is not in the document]
DavidVonDerau: I was intending to hook other things into `show_wireframes`, e.g. we could draw the wireframes around sprites to show the borders of the billboards.

demo/shaders/glsl/pbr_frag.glsl
DavidVonDerau: Done.

demo/src/scenes/shadows_scene.rs
DavidVonDerau: Done.

demo/src/features/mesh/mod.rs
DavidVonDerau: This is renamed to `untextured`. The other one is called `unlit` already.

demo/assets/materials/debug.material
DavidVonDerau: 🤷

rafx-framework/src/render_features/jobs/prepare/submit_node_block.rs
```diff
@@ -34,6 +34,26 @@ impl<SubmitPacketDataT: 'static + Sync + Send + SubmitPacketData>
         }
     }
 
+    pub fn with_capacity_and_optional_feature<
```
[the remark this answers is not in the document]
DavidVonDerau: I've renamed it to `with_capacity_and_feature_flag` and added a comment.

demo/src/assets/gltf/importer.rs
```diff
@@ -283,8 +283,9 @@ impl Importer for GltfImporter {
 
             let mut slot_assignments = vec![];
 
+            let material_data = &material_to_import.asset.material_data;
```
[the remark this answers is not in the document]
DavidVonDerau: Done.

demo/src/lib.rs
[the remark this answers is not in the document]
DavidVonDerau: `show_models` is awkwardly named and I'm not sure what to call it instead. It's actually just skipping the `opaque` and `transparent` render phases entirely.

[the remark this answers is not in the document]
aclysma: For reference, blender has names for four modes:
 * Wireframe
 * Solid (Not exactly "untextured" but seems like a gray that fades to black if a surface normal is ortho to view)
 * Material Preview (it doesn't use lights in the scene, but it's not simply sampling albedo either)
 * Rendered

I think solid or surface would be a good word.

I also think we could slightly change the UX to be:
 * Show Wireframes
 * Show Surfaces
     * Enable Textures
     * Enable Lighting

aclysma:
The only two questions remaining are if we change show_models to something like show_surfaces. Also wondering if 9 PCF samples is a good enough cut. I'd like to see a substantial difference in performance to cut all the way down to 1 sample. 25 -> 9 is already significant.

demo/src/lib.rs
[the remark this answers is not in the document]
DavidVonDerau: image

aclysma:
LGTM, will merge after CI passes

PR #146, opened 2021-05-19
Limited support for `const` in the shader processor

DavidVonDerau:
The shader processor now supports exporting `const` values into the generated Rust code and the usage of `const` values for array sizes.

There are some limitations.

- The shader processor can recognize `*`, `-`, `/`, `+` for the purpose of exporting `const` values but it cannot do so if the const math would be needed to evaluate an array size.
- `const` values that are a `mat3`, `mat4`, `vec3`, or `vec4` will not be exported.

When evaluating an array size, the shader processor will recursively follow `const` values until either a `usize` value is found or one of the above limitations is reached.

### DIFF
--- a/demo/shaders/glsl/mesh_pbr_uniform.glsl
+++ b/demo/shaders/glsl/mesh_pbr_uniform.glsl
@@ -48,6 +48,16 @@ struct ShadowMapCubeData {
     float cube_map_projection_far_z;
 };
 
+const int MAX_POINT_LIGHTS = 16;
+const int MAX_DIRECTIONAL_LIGHTS = 16;
+const int MAX_SPOT_LIGHTS = 16;
+
+const int MAX_SHADOW_MAPS_2D = MAX_DIRECTIONAL_LIGHTS + MAX_SPOT_LIGHTS;
+const int SHADOW_MAP_2D_ARRAY_LEN = 32; // TODO(dvd): Shader processor can't handle const math in array indices.
+
+const int MAX_SHADOW_MAPS_CUBE = MAX_POINT_LIGHTS;
+const int SHADOW_MAP_CUBE_ARRAY_LEN = MAX_POINT_LIGHTS;
+
 // @[export]
 // @[internal_buffer]
 layout (set = 0, binding = 0) uniform PerViewData {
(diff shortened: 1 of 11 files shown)


rafx-shader-processor/src/parse_declarations.rs
```diff
@@ -555,15 +572,94 @@ fn try_parse_binding(code: &[char]) -> Result<Option<ParseBindingResult>, String
...
+            break;
+        } else {
+            fn push_operator(
+                values: &mut Vec<String>,
+                operator: &str,
+            ) {
+                values.push(format!(" {} ", operator));
+            }
+
+            crate::parse_source::skip_whitespace(code, &mut position);
```
DavidVonDerau, line 639: This parsing is actually pretty problematic, and given what it accomplishes, I think it would be cleaner to just copy the entire string between the `=` and the `;`. The issue is that this definition would ban trivial things like using parentheses. 

I think if this was rewritten to just return the string between `=` and `;`, then later code in the pipeline can accommodate that accordingly -- like in this PR, where an array length is simply not allowed to be anything other than a trivial value.

Or, if possible, maybe we could _remove_ the need to actually parse the `usize` for array lengths now that we can embed the `const` itself into the generated Rust code. I'm not sure if anything in the shader processor depends on knowing the actual array length (as a number) vs knowing a symbol (like the name of a `const` identifier).

demo/shaders/glsl/mesh_pbr_uniform.glsl
```diff
@@ -48,6 +48,16 @@ struct ShadowMapCubeData {
     float cube_map_projection_far_z;
 };
 
+const int MAX_POINT_LIGHTS = 16;
+const int MAX_DIRECTIONAL_LIGHTS = 16;
+const int MAX_SPOT_LIGHTS = 16;
+
+const int MAX_SHADOW_MAPS_2D = MAX_DIRECTIONAL_LIGHTS + MAX_SPOT_LIGHTS;
+const int SHADOW_MAP_2D_ARRAY_LEN = 32; // TODO(dvd): Shader processor can't handle const math in array indices.
```
aclysma, line 56: I wonder if these show up in SPIR-V.. if they do, then we might could get the value directly from there. (future work, possibly). spirv-cross might even have something that could help

rafx-shader-processor/src/parse_declarations.rs
aclysma: we may need to know array lengths of certain things like texture arrays so we can flag the binding as being an array of N textures.

aclysma: I think the way this is done now is fine. It's reasonable behavior, not complicated, and we can always build on it later.

rafx-shader-processor/src/parse_source.rs
```diff
@@ -160,9 +202,26 @@ pub(crate) fn try_consume_array_index(
 
         // If this fails, then we may have a string like "123xyz"
         let number: usize = characters_to_string(&code[begin..end]).parse().ok()?;
-
         *position = end;
         Some(number)
+    } else if begin < code.len() {
+        let end = next_non_identifer(code, begin);
+        let identifier = characters_to_string(&code[begin..end]);
```
aclysma, line 209: I think if you have an array float my_array[ABC * XYZ], it will return the value of ABC and ignore the * XYZ part.

Doing this with math operators supported would probably require us to do the arithmetic with correct order of operation (or require explicit parenthesis). As a first pass, maybe we can require that constants are literal, numeric constants.

aclysma (changes requested):
Left some comments, unless I'm mistaken, one of them will be tricky to address.

DavidVonDerau:
Closing due to limitations & a desire to find another approach.

PR #147, opened 2021-05-19
Add egui render feature and switch demo from imgui to egui

aclysma:
* All the existing UI elements use egui now.
 * There used to be an optional "use-imgui" feature. I didn't keep support for it.. the original reason to have it was that the imgui/sdl2 integration manipulated mouse cursors which causes SDL2 to fatal assert on iOS. Now that interop code is in the rafx feature and I can fix that to not try to use mouse cursors next time I run on iOS.
 * Keeping the imgui feature around for now and the feature that turns it on.. we can use it to check that it compiles. I think it's ok for the time being for it to rot a bit. Eventually we'll pull these features into a separate crate or something.
 * I think the linear/sRGB handling for the texture sampling is technically wrong. But the font looks much clearer in this implementation than doing the sRGB conversion. This only affects font rendering, general colors (controlled by vertex attribute) should still get sRGB conversion to linear.

image


demo/src/features/text/jobs/write.rs
```diff
@@ -148,17 +148,6 @@ impl<'write> RenderFeatureWriteJob<'write> for TextWriteJob<'write> {
                         )],
                     )?;
             }
-
-            begin_execute_graph_context
```
DavidVonDerau, line 152: Why did this get removed?

demo/src/features/imgui/internal/imgui_font_atlas.rs
```diff
@@ -23,7 +23,6 @@ impl ImguiFontAtlasData {
 #[derive(Clone)]
 pub struct ImguiFontAtlas(pub ResourceArc<ImageViewResource>);
 
-#[cfg(feature = "use-imgui")]
```
DavidVonDerau, line 26: Should this feature flag have been removed?

DavidVonDerau (approved):
Two minor questions, otherwise looks good.

demo/src/features/imgui/internal/imgui_font_atlas.rs
aclysma: this flag existed because sdl2 + imgui panics on iOS due to it trying to use mouse cursor APIs. Now that the logic for controlling the mouse cursor is in our own crate, we can just make it not do that on iOS. (I'll address it next time I make an iOS build)

demo/src/features/text/jobs/write.rs
aclysma: the lines just above already put the resource into the SHADER_RESOURCE state, so this transition is incorrect and should be removed

PR #155, opened 2021-06-02
Make render thread enable/disable at runtime based on chosen backend

aclysma:
This also moves some Renderer members out of the mutex which cleans up some of the code

### DIFF
--- a/demo/src/features/egui/jobs/extract.rs
+++ b/demo/src/features/egui/jobs/extract.rs
@@ -6,10 +6,11 @@ use rafx::base::resource_map::ReadBorrow;
 use rafx::base::resource_ref_map::ResourceRefBorrowMut;
 use rafx::distill::loader::handle::Handle;
 use rafx::graph::SwapchainSurfaceInfo;
+use rafx::renderer::SwapchainRenderResource;
 
 pub struct EguiExtractJob<'extract> {
     sdl2_egui_manager: TrustCell<ResourceRefBorrowMut<'extract, Sdl2EguiManager>>,
-    swapchain_surface_info: ReadBorrow<'extract, SwapchainSurfaceInfo>,
+    swapchain_surface_info: SwapchainSurfaceInfo,
     asset_manager: ReadBorrow<'extract, AssetManagerRenderResource>,
     egui_material: Handle<MaterialAsset>,
 }
@@ -29,7 +30,11 @@ impl<'extract> EguiExtractJob<'extract> {
                 ),
                 swapchain_surface_info: extract_context
                     .render_resources
-                    .fetch::<SwapchainSurfaceInfo>(),
+                    .fetch::<SwapchainRenderResource>()
+                    .get()
+                    .unwrap()
+                    .swapchain_surface_info
+                    .clone(),
                 asset_manager: extract_context
                     .render_resources
                     .fetch::<AssetManagerRenderResource>(),
@@ -77,6 +82,8 @@ impl<'extract> ExtractJobEntryPoints<'extract> for EguiExtractJob<'extract> {
                 egui_draw_data,
                 egui_material_pass: self
                     .asset_manager
+                    .get()
+                    .unwrap()
                     .committed_asset(&self.egui_material)
                     .unwrap()
                     .get_single_material_pass()
--- a/demo/src/features/egui/jobs/extract.rs
+++ b/demo/src/features/egui/jobs/extract.rs
@@ -1,8 +1,7 @@
 use rafx::render_feature_extract_job_predule::*;
 
 use super::*;
-use rafx::assets::{AssetManagerRenderResource, MaterialAsset};
-use rafx::base::resource_map::ReadBorrow;
+use rafx::assets::{AssetManagerExtractRef, AssetManagerRenderResource, MaterialAsset};
 use rafx::base::resource_ref_map::ResourceRefBorrowMut;
 use rafx::distill::loader::handle::Handle;
 use rafx::graph::SwapchainSurfaceInfo;
@@ -11,7 +10,7 @@ use rafx::renderer::SwapchainRenderResource;
 pub struct EguiExtractJob<'extract> {
     sdl2_egui_manager: TrustCell<ResourceRefBorrowMut<'extract, Sdl2EguiManager>>,
     swapchain_surface_info: SwapchainSurfaceInfo,
-    asset_manager: ReadBorrow<'extract, AssetManagerRenderResource>,
+    asset_manager: AssetManagerExtractRef,
     egui_material: Handle<MaterialAsset>,
 }
 
@@ -37,7 +36,8 @@ impl<'extract> EguiExtractJob<'extract> {
                     .clone(),
                 asset_manager: extract_context
                     .render_resources
-                    .fetch::<AssetManagerRenderResource>(),
+                    .fetch::<AssetManagerRenderResource>()
+                    .extract_ref(),
                 egui_material,
             },
             frame_packet,
--- a/demo/src/features/egui/jobs/extract.rs
+++ b/demo/src/features/egui/jobs/extract.rs
@@ -31,7 +31,6 @@ impl<'extract> EguiExtractJob<'extract> {
                     .render_resources
                     .fetch::<SwapchainRenderResource>()
                     .get()
-                    .unwrap()
                     .swapchain_surface_info
                     .clone(),
                 asset_manager: extract_context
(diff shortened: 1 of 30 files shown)


demo/src/features/egui/jobs/extract.rs
```diff
@@ -77,6 +82,8 @@ impl<'extract> ExtractJobEntryPoints<'extract> for EguiExtractJob<'extract> {
                 egui_draw_data,
                 egui_material_pass: self
                     .asset_manager
+                    .get()
```
DavidVonDerau, line 85: All of the calls to `get().unwrap()` seem the opposite of ergonomic.

DavidVonDerau: Would it be possible to still implement `Deref` and have it automatically do the equivalent of `get().unwrap()`?

rafx-shader-processor/src/lib.rs
```diff
@@ -533,32 +533,40 @@ fn process_glsl_shader(
     // Write out the spv and rust files if desired
     //
     if let Some(spv_file) = &spv_file {
-        std::fs::write(spv_file, output_spv)?;
+        write_output_file(spv_file, output_spv)?;
```
DavidVonDerau, line 536: This commit doesn't need to be included in this PR.

demo/src/features/egui/jobs/extract.rs
aclysma: I don't like it, but I also don't like a Deref that panics. I didn't use just unwrap() because unwrap() generally consumes self. Do you think a different name would be better? Like unwrap_ref(), get_or_panic(), or something like that?

rafx-shader-processor/src/lib.rs
aclysma: I'll squash and merge in order so that they appear as separate changes.

demo/src/features/egui/jobs/extract.rs
DavidVonDerau: What if we had `fetch` return a `ReadBorrow` like before, and added `try_fetch` that would return an `Option<ReadBorrow>` -- does that solve the original problem still? I think that seems more ergonomic since if a resource is required to be present one could use `fetch` without this `get().unwrap()` boilerplate, whereas if it is only conditionally present you use `try_fetch` and bind the `Option` accordingly.

aclysma: Found a different approach that lets us take a "borrow" when creating the extract job.

demo/src/features/skybox/jobs/extract.rs
```diff
@@ -1,15 +1,16 @@
...
 use rafx::distill::loader::handle::Handle;
+use std::marker::PhantomData;
 use std::sync::Arc;
 
 pub struct SkyboxExtractJob<'extract> {
-    asset_manager: ReadBorrow<'extract, AssetManagerRenderResource>,
+    asset_manager: AssetManagerExtractRef,
     skybox_material: Handle<MaterialAsset>,
     skybox_texture: Handle<ImageAsset>,
+    phantom_data: PhantomData<&'extract ()>,
```
DavidVonDerau, line 13: Is this `phantom_data` needed? Other jobs don't have it.

demo/src/features/imgui/jobs/extract.rs
```diff
@@ -6,10 +6,11 @@ use rafx::base::resource_map::ReadBorrow;
 use rafx::base::resource_ref_map::ResourceRefBorrowMut;
 use rafx::distill::loader::handle::Handle;
 use rafx::graph::SwapchainSurfaceInfo;
+use rafx::renderer::SwapchainRenderResource;
 
 pub struct ImGuiExtractJob<'extract> {
     sdl2_imgui_manager: TrustCell<ResourceRefBorrowMut<'extract, Sdl2ImguiManager>>,
-    swapchain_surface_info: ReadBorrow<'extract, SwapchainSurfaceInfo>,
+    swapchain_surface_info: SwapchainSurfaceInfo,
     asset_manager: ReadBorrow<'extract, AssetManagerRenderResource>,
```
DavidVonDerau, line 14: Should this be `AssetManagerExtractRef`?

rafx-renderer/src/swapchain_render_resource.rs
```diff
@@ -37,11 +44,21 @@ impl SwapchainResources {
...
+        });
+
+        Ok(())
+    }
+
+    pub fn clear_swapchain(&mut self) {
+        self.0 = None;
+    }
+
+    pub fn get(&self) -> Option<&SwapchainRenderResourceInner> {
```
DavidVonDerau, line 61: Can we rename this to `try_get` and add a separate `get` that performs the unwrap, then change the current uses of `try_get` to the `get` call instead?

demo/src/features/tile_layer/jobs/extract.rs
```diff
@@ -1,14 +1,15 @@
...
+use rafx::assets::{AssetManagerExtractRef, AssetManagerRenderResource, MaterialAsset};
 use rafx::distill::loader::handle::Handle;
+use std::marker::PhantomData;
 
 pub struct TileLayerExtractJob<'extract> {
-    asset_manager: ReadBorrow<'extract, AssetManagerRenderResource>,
+    asset_manager: AssetManagerExtractRef,
     tile_layer_material: Handle<MaterialAsset>,
     render_objects: TileLayerRenderObjectSet,
+    phantom_data: PhantomData<&'extract ()>,
```
DavidVonDerau, line 12: Is this `phantom_data` needed?

rafx-renderer/src/render_frame_job.rs
```diff
@@ -13,7 +12,9 @@ pub struct RenderFrameJobResult;
 /// The `RenderFrameJob` is responsible for the `prepare` and `write` steps of the `Renderer` pipeline.
 /// This is created by `Renderer::try_create_render_job` with the results of the `extract` step.
 pub struct RenderFrameJob {
-    pub renderer: Renderer,
+    //pub renderer: Renderer,
```
DavidVonDerau, line 15: Commented out `Renderer`

rafx-assets/src/assets/asset_manager_render_resource.rs
```diff
@@ -1,23 +1,46 @@
...
+    pub fn end_extract(&mut self) {
+        self.0
+            .take()
+            .to_owned()
+            .expect("Reference to AssetManager is still in use");
+    }
+
+    /// Returns an extract ref. Panics if called while not extracting. The ref must be dropped
+    /// before end_extract() is called.
+    pub fn extract_ref(&self) -> AssetManagerExtractRef {
```
DavidVonDerau, line 31: Maybe rename this `get` to match swapchain? And we could have `try_get` that would return an option.

rafx-renderer/src/renderer.rs
```diff
@@ -368,15 +362,17 @@ impl Renderer {
 
         asset_manager.on_begin_frame()?;
 
-        render_resources.insert(swapchain_surface_info.clone());
+        //render_resources.insert(swapchain_surface_info.clone());
```
DavidVonDerau, line 365: Commented code

rafx-renderer/src/swapchain_handling.rs
```diff
@@ -127,20 +125,20 @@ impl<'a> RafxSwapchainEventListener for SwapchainHandler<'a> {
...
+        let mut swapchain_render_resource = self
+            .renderer
+            .render_resources
+            .fetch_mut::<SwapchainRenderResource>();
+        swapchain_render_resource.clear_swapchain();
 
         //TODO: Explicitly remove the images instead of just dropping them. This prevents anything
         // from accidentally using them after they've been freed
-        //swapchain_resources.swapchain_images.clear();
+        //swapchain_render_resource.swapchain_images.clear();
```
DavidVonDerau, line 138: Can this commented code be removed or is the `TODO` still relevant?

demo/src/features/skybox/jobs/extract.rs
aclysma: it's needed because the 'extract lifetime was unused. There may be a way to remove the lifetime, but I felt that consistency/ease of adding it back was better.

rafx-renderer/src/swapchain_render_resource.rs
aclysma: yeah

demo/src/features/tile_layer/jobs/extract.rs
aclysma: same as above

rafx-renderer/src/swapchain_handling.rs
aclysma: I think this is still worth doing at some point

rafx-assets/src/assets/asset_manager_render_resource.rs
aclysma: I'm not sure about using get() for something that is returning anything more than a reference to a wrapped thing

DavidVonDerau (approved):

PR #156, opened 2021-06-02
Remove compute test

aclysma:
I don't think this is useful to have around right now and it's blocking the GLES3 backend from working with the demo.

(We can always bring this back later if we want!)


DavidVonDerau:
I would approve this PR if it was _just_ the compute commit, but it also includes several other PRs in it.

DavidVonDerau (changes requested):
Minimize this PR to just the compute removal.

DavidVonDerau (approved):

PR #159, opened 2021-06-04
Remove usage of std::time::Instant

aclysma:
Also remove other usages of time-based abstractions that aren't available in wasm

### DIFF
--- a/demo/src/time.rs
+++ b/demo/src/time.rs
@@ -1,6 +1,9 @@
+use rafx::base::Instant;
+use std::time::Duration;
+
 /// Records time when created and logs amount of time passed when dropped
 pub struct ScopeTimer<'a> {
-    start_time: std::time::Instant,
+    start_time: Instant,
     name: &'a str,
 }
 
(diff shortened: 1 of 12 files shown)


DavidVonDerau (approved):

demo/src/time.rs
```diff
@@ -90,10 +93,6 @@ impl TimeResource {
         }
     }
 
-    pub fn system_time(&self) -> &TimeContext {
```
DavidVonDerau, line 93: I wonder if it's worth making this `Option<TimeContext>` rather than removing it entirely.

demo/src/time.rs
```diff
@@ -195,19 +189,14 @@ impl TimeState {
...
         // Determine length of time since last tick
-        let now_instant = time::Instant::now();
+        let now_instant = Instant::now();
         let elapsed = now_instant - self.previous_update_instant;
         self.previous_update_instant = now_instant;
         self.app_time_context.update(elapsed);
     }
 
-    /// System time that the application started
-    pub fn app_start_system_time(&self) -> &time::SystemTime {
```
DavidVonDerau, line 205: Ditto here -- same as previous question.

demo/src/time.rs
aclysma: This is from an older project that was more "engine"-like and we're just using it as demo code. So I think if we don't use it, there isn't any reason to keep it around.

PR #161, opened 2021-06-04
Dependency updates

aclysma:
- Remove image2 dependency and use image everywhere instead
- Update to latest gltf
- Make basis-universal optional
- Make egui optional
- Update to more recent rand/pcg_rand version

These changes will be helpful for supporting building in wasm

### DIFF
--- a/demo/Cargo.toml
+++ b/demo/Cargo.toml
@@ -9,7 +9,7 @@ publish = false
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [dependencies]
-rafx = { path = "../rafx", features = ["renderer"] }
+rafx = { path = "../rafx", features = ["renderer", "basis-universal"] }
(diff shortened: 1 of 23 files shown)


demo/Cargo.toml
```diff
@@ -61,13 +59,15 @@ bevy_tasks = "0.5.0"
 default = [
     "sdl2-bundled",
     "sdl2-static-link",
-    "profile-with-puffin",
-    #"profile-with-stats-alloc", # Cannot be enabled with "profile-with-tracy-memory". Will run renderer / game loop single-threaded.
+    #"profile-with-puffin",
```
DavidVonDerau, line 62: Is it necessary for `profile-with-puffin` to be removed from the default options?

DavidVonDerau (approved):

demo/Cargo.toml
aclysma: No, good point. :)

PR #162, opened 2021-06-06
Switch demo to winit

aclysma:
While I do strongly prefer SDL2 in my own projects, winit is more wasm-friendly and may be easier for some people to build/run. The examples still use SDL2 so support for it won't rot.

### DIFF
--- a/demo/Cargo.toml
+++ b/demo/Cargo.toml
@@ -14,9 +14,7 @@ shaders = { path = "shaders" }
 distill = { version = "=0.0.3", features = ["serde_importers"] }
 uuid = "0.8"
 type-uuid = "0.1"
-sdl2 = { version = "0.34", features = ["raw-window-handle"] }
 imgui = { version = "0.7", optional = true }
-imgui-sdl2 = { version = "0.14.0", optional = true }
 egui = { version = "0.12", features = ["default_fonts"], optional = true }
 copypasta = { version = "0.7", optional = true }
 puffin_egui = { version = "0.2.0", optional = true }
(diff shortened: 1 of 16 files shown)


demo/Cargo.toml
```diff
@@ -54,28 +52,29 @@ ron = "0.6"
...
+env_logger = "0.6"
+winit = { version = "0.24" }
 
 [features]
 default = [
-    "sdl2-bundled",
-    "sdl2-static-link",
     "profile-with-puffin",
-    "egui"
+    "use-egui"
```
aclysma, line 64: conventionally it should be possible to use the egui crate name directly to turn on egui related stuff, but we need it to turn on multiple crates, so I think we have to use a feature

DavidVonDerau (approved):
Looks good except for the failing build.

demo/Cargo.toml
```diff
@@ -14,10 +14,9 @@ shaders = { path = "shaders" }
 distill = { version = "=0.0.3", features = ["serde_importers"] }
 uuid = "0.8"
 type-uuid = "0.1"
-sdl2 = { version = "0.34", features = ["raw-window-handle"] }
 imgui = { version = "0.7", optional = true }
-imgui-sdl2 = { version = "0.14.0", optional = true }
 egui = { version = "0.12", features = ["default_fonts"], optional = true }
+copypasta = { version = "0.7", optional = true }
```
DavidVonDerau, line 19: It looks like this transitively included `lazy-bytes-cast` which was BPL. 
image

DavidVonDerau: This is also raising another error in `cargo-deny`
image

DavidVonDerau: `xcb` is included by `copypasta`

aclysma:
I switched it to use clipboard and turned it off by default as there is a security advisory for the replacement. We could consider removing in the future. The advisory only affects x11. I didn't see a good way to disable it for just x11.

PR #165, opened 2021-06-15
Add Features to Docs.rs & Inline Re-exported Crates

zicklag:
Also updates some presumably out-of-date docs.

### DIFF
--- a/rafx/Cargo.toml
+++ b/rafx/Cargo.toml
@@ -79,3 +79,6 @@ required-features = ["framework"]
 name = "asset_triangle"
 path = "examples/asset_triangle/asset_triangle.rs"
 required-features = ["assets"]
+
+[package.metadata.docs.rs]
+features = ["rafx-vulkan", "framework", "assets", "renderer"]
--- a/rafx/src/lib.rs
+++ b/rafx/src/lib.rs
@@ -1,18 +1,19 @@
-//! NOTE: **docs.rs may not generate complete documentation for rafx because it does not enable any
-//! features in cargo when building the docs.** To generate complete documentation locally, run
-//! `cargo doc --no-deps --open` in the root of the crate.
-//!
-//! **Please see [additional documentation here](https://github.com/aclysma/rafx/blob/master/docs/index.md)**
+//! NOTE: **docs.rs may not generate complete documentation for rafx because it does not enable all
+//! of the features in cargo when building the docs.** To generate complete documentation locally,
+//! run `cargo doc --no-deps --open` in the root of the crate.
 //!
 //! Rafx is a multi-backend renderer that prioritizes performance, flexibility, and productivity. It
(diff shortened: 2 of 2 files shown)


aclysma:
The cargo.toml change is fine. For the lib.rs:
 * I prefer the flat module layout (For a project in this phase of maturity I usually avoid renames/moves unless it's a substantial improvement in clarity. I think we might just go in circles otherwise.)
 * I would keep the first doc comment explaining how to generate full docs. (Even with the feature flags in the cargo.toml, it's still not really complete since the end-user can pick other docs.)
 * What does the doc(inline) do? Again for this phase of the project, I'm not too concerned with what rustdoc looks like, and default behavior I saw previously seemed fine. I'm open to this change, just again for a project that is still not mature and likely to change, I err towards avoiding purely cosmetic changes unless one way is clearly better than another.

aclysma:
Just looked up doc(inline). I think the way it presents in rustdoc now is fine (but if you want to paste a screenshot of what it would look like after the change, happy to look with an open mind!)

FYI while rafx-api did get a pretty decent pass on rustdoc (as I think `rafx-api` is pretty mature and unlikely to change significantly, there is little to no rustdoc comments in the higher-level code, and I don't plan on making a pass to improve this until it is more mature.

zicklag:
> I prefer the flat module layout (For a project in this phase of maturity I usually avoid renames/moves unless it's a substantial improvement in clarity. I think we might just go in circles otherwise.)

Makes sense. 👍

> I would keep the first doc comment explaining how to generate full docs. (Even with the feature flags in the cargo.toml, it's still not really complete since the end-user can pick other docs.)

There I put that back in there.I'll put that back in there.

> Just looked up doc(inline). I think the way it presents in rustdoc now is fine (but if you want to paste a screenshot of what it would look like after the change, happy to look with an open mind!)

Here's with the inlined docs:

image

And here's without:

image

It's subjectively "nicer" looking without the re-exports section and you also get the descriptions of the re-exported modules. For instance, with the `framework` module, you can see the description of it in the inlined version, but you would have to click on it in the re-exports section.

The biggest reason to inline the docs, though, is for searching. If you search `RafxApi` on docs.rs/rafx right now it will not find anything because `RafxApi` is in the `api` re-export. If you inline the docs, though, you can search `RafxApi` and it will it in `rafx::api::RafxApi` because it treats `api` like an inline module instead of a separate crate.

> FYI while rafx-api did get a pretty decent pass on rustdoc (as I think rafx-api is pretty mature and unlikely to change significantly, there is little to no rustdoc comments in the higher-level code, and I don't plan on making a pass to improve this until it is more mature.

OK, sounds good.

aclysma:
Sorry for the delay on this, I've been busy with prototyping a blender export workflow for models and scenes and not looking at GitHub, and I forgot about this :)

PR #167, opened 2021-06-15
Use Fully-Qualified Type Paths in Macros

zicklag:
I ran into an error because I didn't `use` the requried types before calling the macro. This makes sure that you don't have to `use` anything and the macro should work regardless.

Also updated an out-of-date code sample in one of the doc readmes.


aclysma:
This was rolled up into https://github.com/aclysma/rafx/commit/4de391a69babf8de1db5e815010e17eadf732b4c

PR #170, opened 2021-06-30
Enable features on docs.rs

jamen:
I think this should work, see [this](https://docs.rs/about/metadata). I wasn't sure about backends, so I left them out, but `all-features = true` might be nice.


jamen:
Oh oops #165

PR #184, opened 2021-11-02
Acknowledge `wgpu` relicensing to Apache-2.0/MIT

tasgon:
`wgpu-rs` was relicensed under Apache/MIT [roughly 4 months ago](https://github.com/gfx-rs/wgpu/commit/74d2becb1856162828122cbf8f666a90861f3d71). This PR adjusts the comparison page to reflect that change.


aclysma:
Thanks, I’ve been meaning to rewrite that page, not just to include the license change but to reflect that gfx-hal is discontinued and that wgpu’s crate layout is much simpler now. These were all factors that made me avoid using or recommending wgpu, but now I would actually recommend wgpu over rafx-api for some projects/use-cases. I’ll try to make these changes this week. (Since it’s prose with some opinion, I’d like to make the edits myself in this case.) But thanks for the reminder. I’ll leave this open until I update that page.

aclysma:
Doc has been updated to reflect gfx-hal being deprecated, and wgpu being relicensed and restructured to be simpler than it was in late 2020.

PR #196, opened 2021-12-04
Max color component value

aclysma:


aclysma:
I merged this over manually in 3d523ee8b0118aef9594e08aafb7fb1d75383b0f so that the two main changes in the branch could be separate commits

Issue #223, opened 2022-02-01
crash when running demo in Linux


BillyDM:
I'm getting a crash when trying to run the demo in Linux with the vulkan backend. This is the backtrace:
```
thread 'main' panicked at 'index out of bounds: the len is 3 but the index is 3', rafx-plugins/src/pipelines/modern/graph_generator/mod.rs:254:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:107:14
   2: core::panicking::panic_bounds_check
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:75:5
   3: rafx_plugins::pipelines::modern::graph_generator::generate_render_graph
   4: <rafx_plugins::pipelines::modern::plugin::ModernPipelineRendererPlugin as rafx_renderer::renderer_pipeline_plugin::RendererPipelinePlugin>::generate_render_graph
   5: rafx_renderer::renderer::Renderer::start_rendering_next_frame
   6: demo::DemoApp::update
   7: demo::update_loop::{{closure}}
   8: winit::platform_impl::platform::x11::EventLoop<T>::run
   9: winit::platform_impl::platform::EventLoop<T>::run
  10: winit::event_loop::EventLoop<T>::run
  11: demo::update_loop
  12: demo::main_native::main_native
  13: demo::main
```

I'm running an up-to-date Manjaro Linux environment with the Budgie DE, and I'm running an AMD RX570 GPU.

aclysma:
I wasn't able to reproduce a crash but I did trigger some validation errors in vulkan/windows 10 when I switched to a 2d scene. I probably introduced this recently. In case this is something different, if you're willing to help find where this is coming from, could you capture a more detailed stack by using the environment args `RUST_BACKTRACE=full` in debug mode? The odd thing about this stack is that I wouldn't expect this logic to differ across platforms or graphics APIs.

Also, could you confirm if this happened immediately on launch or if you had to change scenes (arrow keys.)

(While I know people have run rafx successfully in linux, I personally only use macOS/windows, so I may have a hard time helping much with linux issues)

BillyDM:
[...]

PR #224, opened 2022-02-06
Avoid 1D group counts >65535

aclysma:


aclysma:
This may help with the first crash reported in #223

PR #240, opened 2022-09-10
Add support for debug naming and debug labels, automatically name nodes in the render graph

RDambrosio016:
This PR adds APIs for naming resources (currently just buffers and textures) and command buffer labels in rafx-api. The rendergraph uses the APIs to label nodes if they have a name, which significantly improves the debugging experience: 

unknown (16)

It currently only supports vulkan, other backends have the functions but they are no-ops. But adding support for other platforms should be really easy, i just have no way to test it for backends like metal.

### DIFF
--- a/rafx-api/src/backends/gles2/device_context.rs
+++ b/rafx-api/src/backends/gles2/device_context.rs
@@ -79,6 +79,7 @@ impl RafxDeviceContextGles2Inner {
 
         let device_info = RafxDeviceInfo {
             supports_multithreaded_usage: false,
+            supports_debug_names: false,
             min_uniform_buffer_offset_alignment: pack_alignment,
             min_storage_buffer_offset_alignment: pack_alignment,
             upload_buffer_texture_alignment: pack_alignment,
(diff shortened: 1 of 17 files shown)


RDambrosio016:
@aclysma how would u prefer i fix the clippy complaints? should i just `not(feature = "rafx-vulkan")` the pattern? i added a fallback so it would work on other backends

aclysma:
I would prefer to mirror the other functions that have match cases for all backends. BTW I wanted to play with this change a bit for myself so I have a branch based on this PR that includes that change. (https://github.com/aclysma/rafx/tree/debug-names) I also wanted to try some ideas for the public APIs and implement this for the metal backend. I plan to do a second PR that includes your commits and we can discuss.

RDambrosio016:
No problem, take your time

aclysma:
Got it basically working in metal. Works with the labels you set up in command buffers and I added some on some buffers/textures.

image

aclysma:
Started a new PR that includes these changes and adds support for metal in #243

PR #254, opened 2023-07-05
feat: prellocate-semaphores-fences

pollend:
it makes sense to reserve more framebuffers then anyone would ever use and it should avoid an extra indirection for all of these elements.

### DIFF
--- a/rafx-api/Cargo.toml
+++ b/rafx-api/Cargo.toml
@@ -16,6 +16,7 @@ categories = ["game-development", "game-engines", "graphics", "multimedia", "ren
 [dependencies]
 rafx-base = { version = "0.0.15", path = "../rafx-base" }
 
+arrayvec = "0.7.4"
 log = "0.4"
 crossbeam-channel = "0.5"
 fnv = "1.0"
--- a/rafx-api/src/extra/swapchain_helper.rs
+++ b/rafx-api/src/extra/swapchain_helper.rs
@@ -6,6 +6,7 @@ use crate::{
 use crossbeam_channel::{Receiver, Sender};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
+use arrayvec::ArrayVec;
 
 /// May be implemented to get callbacks related to the swapchain being created/destroyed. This is
 /// optional.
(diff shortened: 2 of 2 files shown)


aclysma:
I'm not sure it's worth taking a dependency for this. The cost of this indirection is dominated by actually using the fences/semaphores.

pollend:
I though you planned to use smallvector

aclysma:
I'm not strongly against the change. It also occurs to me filling out structures for vulkan/dx12 may be less hazardous with something like arrayvec/smallvec. It's very easy to accidentally pass a pointer to the contents of a Vec to DX12/vulkan structures and do something that reallocates the Vec making that pointer invalid.

I think there's several smallvec/arrayvec implementations, but I haven't taken the time to review them. Is there a reason you like this one in particular? Just curious.

In the cases where we've had a lot of allocations we ended up using AtomicOnceCellArray/AtomicOnceCellStack. This tends to be when copying data from the "game" to the render thread, so it's per-frame allocations. But I think if we do want to change this to avoid the indirection, something like arrayvec/smallvec would be the better choice.

I'm still not sure that this is really the place where arrayvec/smallvec is going to make much of a difference. But if you want to move forward with the change I'm ok with it. I would just ask that 1) not use the "8" magic number in multiple places 2) We should pin this to MAX_NUM_FRAMES_IN_FLIGHT + 1. There's no reason to have a sync primitive per image. (On that note maybe we can use a simple array.)

pollend:
keep the changes small the focused, are there key areas of allocation to focus on?

aclysma:
> are there key areas of allocation to focus on?

The areas we (actualy mostly @DavidVonDerau) put a lot of work into were capturing from the main thread to the rendering thread. You could find these by searching AtomicOnceCellArray/AtomicOnceCellStack. These were more-than-per-frame, dynamically growing allocations that definitely show up as small latency spikes. Something like this where we are very rarely re-allocating, and it's along with a lot of other work that's heavy (like rebuilding a swapchain) is not going to be noticeable.

It's true this is one less indirection and we access the data per-frame, but given this is a "volunteer" project, people's time is at a premium. I think there are other places where we get more impact like 1) ways to make the renderer scale better (i.e. number of lights/meshes/etc.) 2) ways to make the image quality better (i.e. improved rendering techniques) 3) making things easier to use either by tools or APIs.

Again this isn't a bad change, I just think in terms of impact I think there are other places that add more value for the time spent.

The projects I am considering taking on at some point:
 * Finish bringing up the DX12 backend. It mostly works!
 * Switch from GLSL to HLSL (GLSL is basically dead/not getting updates at this point, and dxc gets better and better. All the industry investment is there right now and it's being done mostly in the open.)
 * New content pipeline to address some of the usability issues with distill
 * Deferred texturing
 * Hybrid ray-tracing (i.e. raytraced shadows/reflections)

Also just a heads up I will probably be shifting my focus with this library from "general purpose do anything" to "platform for graphics research" as this is what I personally need. So less focus on baseline compatibility everywhere and more focus on modernization/graphics features.

